PR #84, opened 2019-03-07
Fix setup

xshill:
This should fix our installation woes for virtual environments.


xshill:
@LacroixEtienne if / when you have time, can you check if the Dockerfile works properly? Docker breaks my internet every time I try to run or build a container.

PR #102, opened 2019-04-22
Added some logging of X224 negotiation failures

obilodeau:
WIP because I left a TODO task in there and forgot why.

Please review and enlighten me.


pyrdp/enum/rdp.py
```diff
@@ -347,6 +347,28 @@ class NegotiationProtocols(IntFlag):
     EARLY_USER_AUTHORIZATION_RESULT = 0b00001000
 
 
+class NegotiationFailureCode(IntEnum):
```
Res260: I like when we write the source for those flags in the class' documentation. In this case its https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/1b3920e7-0116-4345-bc45-f2c4ad012761

pyrdp/mitm/X224MITM.py
```diff
@@ -99,6 +99,12 @@ def onConnectionConfirm(self, _: X224ConnectionConfirmPDU):
         protocols = NegotiationProtocols.SSL if self.originalRequest.tlsSupported else NegotiationProtocols.NONE
 
         parser = NegotiationResponseParser()
+        response = parser.parse(pdu.payload)
+        if response.type == NegotiationType.TYPE_RDP_NEG_FAILURE:
```
Res260: We took the habit to make these kind of checks with `isinstance(response, NegotiationFailurePDU)` whenever possible, since it allows IDEs with content-aware linting to know that it can assume that the `response` variable is indeed a `NegotiationFailurePDU` in this `if` block. 

With this current code, Pycharm would lint an error since it can't assume that `response` contains the `failureCode` attribute. I think it would also have spotted that the `NegotiationResponseParser.parse()` method has a return type of `NegotiationResponsePDU` and that the comparison is conflicting since the `NegotiationFailurePDU` does not inherit `NegotiationResponsePDU`

pyrdp/parser/rdp/negotiation.py
```diff
@@ -93,8 +93,13 @@ def parse(self, data: bytes) -> NegotiationResponsePDU:
...
             length = Uint16LE.unpack(stream)
-            selectedProtocols = Uint32LE.unpack(stream)
-            return NegotiationResponsePDU(type, flags, selectedProtocols)
+
+            if type == NegotiationType.TYPE_RDP_NEG_FAILURE:
+                failureCode = Uint32LE.unpack(stream)
+                return NegotiationFailurePDU(type, flags, failureCode)
+            else:
+                selectedProtocols = Uint32LE.unpack(stream)
+                return NegotiationResponsePDU(type, flags, selectedProtocols)
```
Res260: As I said in my previous comment, I think this should be changed because the return type isnt telling the truth now. Since the only difference is that selectedProtocol == failureCode for the FailurePDU, maybe make `NegotiationFailurePDU` inherit `NegotiationResponsePDU` and add something like `failureCode = selectedProtocols` in the constructor? @xshill what do you think?

pyrdp/mitm/X224MITM.py
```diff
@@ -99,6 +99,12 @@ def onConnectionConfirm(self, _: X224ConnectionConfirmPDU):
...
+            self.log.info("The server failed the negotiation. Error: %(error)s", {"error": NegotiationFailureCode.getMessage(response.failureCode)})
+            return
+
+        # TODO: cleanup
```
Res260: Maybe you meant exit the program differently? Simply returning in this scenario seems odd 🤔

Res260:
Press F to pay respects to the forgotten PR 🙏

pyrdp/parser/rdp/negotiation.py
Res260: Its a bad idea

pyrdp/mitm/X224MITM.py
obilodeau: Fixed

pyrdp/mitm/X224MITM.py
obilodeau: Yeah.. as discussed, we forgot and the change has is should be beneficial so let's go ahead.

obilodeau:
After an in-person review refresh with Emilio, I made all the requested changes.

Please consider for merge again. May I suggest using a squash merge :wink:

Res260:
Will test then merge if nothing is on fire!

Res260:
Tested the changes, and added code to transfer the FailurePDU back to the client so the client can gracefully exit with a message saying that NLA is not supported on its computer (lol). Merging.

Issue #103, opened 2019-05-08
Harvest credentials entered on Windows lock screen


Pourliver:
Hey ! :wave: 

Currently, PyRDP logs any credentials sent by the RDP client when initializing the connection. The builtin RDP clients in Windows 7 and up won't send any username / password in the initialization by default. They can be configured to send a username, but not a password. In that scenario, the user proceeds to manually enter his credentials in the Windows lock screen, and it is not logged by PyRDP.

Even though we can still get the username and password by watching the replay, via the keylogger, I think it would be a great addition to automatically scrape the keylogger and harvest the username and password from the lock screen.

I'll try a proof of concept tomorrow. If you have any suggestions please let me know.

Pourliver:
Without going into implementation details, I have a basic keylogger that prints the first 3 input whenever a client connects. Here is a very filtered output showcasing this feature. 

![Screenshot from 2019-05-09 15-36-38](https://user-images.githubusercontent.com/11776687/57481533-5802e880-7270-11e9-9e24-5c9386e1043b.png)

I plan on making the number of keylogged input configurable, and default to 2. This keylogger doesn't handle backspaces, since the user may have to fill both the username and password field. 

Correct me if I'm wrong, but I haven't noticed any way to know if the user passed the lock screen. This implies that we won't be able to reliably steal the user input 100% of the time, since the user may enter wrong credentials a few times.

xshill:
Seems like a neat feature. For detecting log ins, look at the message sent through the RDPDR channel when the user is logged on (I talked about it in this issue: https://github.com/GoSecure/pyrdp/issues/98).

Res260:
^ However, if the RDPDR channel doesnt exist, it won't get sent, so keep that in mind :)

Pourliver:
Yeah I've seen that issue, but I was not too sure if it was reliable. Since it is used by default in WIndows, I'll print the last input before "PAKID_CORE_USER_LOGGEDON". If the channel never exists, then I think it would be fine if the credentials are never printed.

Pourliver:
Implemented in #106

Issue #98, opened 2019-04-15
Use RDPDR logged on message to time payloads properly


xshill:
The RDPDR channel seems to send a special PDU when the user has logged on. This would be ideal for timing on-connection payloads.

The packet ID for this message is PAKID_CORE_USER_LOGGEDON.

To do:

1. Check when the message is sent. It's possible we might have to use a small delay to make sure everything works properly. Also check if the message is always sent if the RDPDR channel is initialized, even if the client doesn't actually use the channel.
2. Run the payload when receiving the message in the cases where we can predict that the message will indeed be sent. In other cases, keep it the way it currently works (user-configured delay).

Res260:
Need to confirm, but i'm 90% sure the message does not get sent if the RDPDR channel is not activated. The purpose of this PDU is to prevent the client from sending useless disk/other IO requests that won't get answered otherwise

Res260:
However, in a typical work environnent, I think its a fair assumption to say that the channel is active most of the time (since I think its activated by default)

xshill:
What I would really like to know is whether or not it gets sent even if the client does not have the intention of mapping drives or anything. I'm think all MS clients initialize all channels anyway. If this is indeed the case, then we could probably make use of that message 100% of the time.

Res260:
>It is sent when the user is logged on, or when the server reconnects to a session with an already logged-on user, immediately after it receives a Client Core Capability Response. This packet is sent only if the client's general capability set has the RDPDR_USER_LOGGEDON_PDU bit set in the extendedPDU field.

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/2ecd5c18-8c8c-4f5d-97ef-794ce19e91ba

So the client needs to send its capabilities first, we just gotta confirm that tsclient.exe indeed does that even if you don't map anything (I'd think its the case)

Res260:
Other reference: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/006b6528-0684-438e-8c62-e30c8c27f0f0

Pourliver:
I've done a few tests to understand how the RDPDR channel works, and from my understanding, it only seems enabled when connecting via the default Windows RDP client.

I've tried a few other RDP clients. For example, using Windows Remote Client, the server sends the "PAKID_CORE_SERVER_ANNOUNCE" request after connection, but the client never sent any DeviceRedirectionPDU.

alxbl:
[...]

PR #106, opened 2019-05-13
Added credential logger

Pourliver:
Here is the password harvester showcased in this issue #103.

I implemented a new logger based on the FastPathLogger. The new CredentialLogger is a singleton, since it needs to be referenced from somewhere it won't be defined in order to print credentials. I tried an implementation with a static method and a few static variable at first, but this once made more sense since there will never be two instances of a logger.

We keylog everything the users sends, and we save them as a candidate whenever the user presses the "ENTER" key. When we receive a confirmation that the user logged in, we print the candidate (or the buffer if the user submitted the form with a click)

When the user logs in, the logger deactivate itself. Whenever a new client connects, it gets reactivated.

This should give us good credentials most of the time, but It may give false-positives if the user press keys while not in an input field, or if he spams his keyboard right after being logged in but before receiving the "RDPDR" signal.

### DIFF
--- a/pyrdp/mitm/DeviceRedirectionMITM.py
+++ b/pyrdp/mitm/DeviceRedirectionMITM.py
@@ -11,12 +11,13 @@ from pathlib import Path
 from typing import Dict, Optional, Union
 
 from pyrdp.core import FileProxy, ObservedBy, Observer, Subject
-from pyrdp.enum import CreateOption, DeviceType, DirectoryAccessMask, FileAccessMask, FileAttributes, \
+from pyrdp.enum import CreateOption, DeviceRedirectionPacketID, DeviceType, DirectoryAccessMask, FileAccessMask, FileAttributes, \
     FileCreateDisposition, FileCreateOptions, FileShareAccess, FileSystemInformationClass, IOOperationSeverity, \
     MajorFunction, MinorFunction
 from pyrdp.layer import DeviceRedirectionLayer
 from pyrdp.mitm.config import MITMConfig
 from pyrdp.mitm.FileMapping import FileMapping, FileMappingDecoder, FileMappingEncoder
+from pyrdp.mitm.state import RDPMITMState
 from pyrdp.pdu import DeviceAnnounce, DeviceCloseRequestPDU, DeviceCloseResponsePDU, DeviceCreateRequestPDU, \
     DeviceCreateResponsePDU, DeviceDirectoryControlResponsePDU, DeviceIORequestPDU, DeviceIOResponsePDU, \
     DeviceListAnnounceRequest, DeviceQueryDirectoryRequestPDU, DeviceQueryDirectoryResponsePDU, DeviceReadRequestPDU, \
@@ -53,7 +54,7 @@ class DeviceRedirectionMITM(Subject):
     FORGED_COMPLETION_ID = 1000000
 
 
-    def __init__(self, client: DeviceRedirectionLayer, server: DeviceRedirectionLayer, log: LoggerAdapter, config: MITMConfig):
+    def __init__(self, client: DeviceRedirectionLayer, server: DeviceRedirectionLayer, log: LoggerAdapter, config: MITMConfig, state: RDPMITMState):
         """
         :param client: device redirection layer for the client side
         :param server: device redirection layer for the server side
@@ -64,6 +65,7 @@ class DeviceRedirectionMITM(Subject):
 
         self.client = client
         self.server = server
+        self.state = state
         self.log = log
         self.config = config
         self.currentIORequests: Dict[int, DeviceIORequestPDU] = {}
@@ -127,6 +129,10 @@ class DeviceRedirectionMITM(Subject):
 
         elif isinstance(pdu, DeviceListAnnounceRequest):
             self.handleDeviceListAnnounceRequest(pdu)
+
+        elif isinstance(pdu, DeviceRedirectionPDU):
+            if pdu.packetID == DeviceRedirectionPacketID.PAKID_CORE_USER_LOGGEDON:
+                self.handleClientLogin()
 
         if not dropPDU:
             destination.sendPDU(pdu)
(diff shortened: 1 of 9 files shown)


Res260:
@xshill Been a while since I dove into the code, but wasnt there a way to avoid having a singleton for this?

Pourliver:
Thanks a lot for the feedback! I lacked the vision to clearly see how to correctly integrate this feature in the project, but your proposal makes a lot of sense.

Pourliver:
I refactored the feature according to your feedback. It's much cleaner, and we even have a good base to time payloads when user logs in (as requested in #98).

pyrdp/mitm/FastPathMITM.py
```diff
@@ -37,6 +37,34 @@ def onClientPDUReceived(self, pdu: FastPathPDU):
         if self.state.forwardInput:
             self.server.sendPDU(pdu)
 
+        if not self.state.loggedIn:
+            for event in pdu.events:
+                if isinstance(event, FastPathScanCodeEvent):
+                    self.onScanCode(event.scanCode, event.isReleased, event.rawHeaderByte & 2 != 0)
```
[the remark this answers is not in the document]
Pourliver: I copied / pasted that part of the code from somewhere else in the code base. I can't say I've looked much into it. Will do tomorrow.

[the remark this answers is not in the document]
Res260: ayy lmao

pyrdp/mitm/DeviceRedirectionMITM.py
```diff
@@ -270,6 +276,22 @@ def handleCloseResponse(self, request: DeviceCloseRequestPDU, _: DeviceCloseResp
             self.saveMapping()
 
 
+    def handleClientLogin(self):
+        """
+        Handle events that should be triggered when a client logs in.
+        """
+
+        if self.state.candidate or self.state.inputBuffer:
+            self.log.info("Credentials candidate: %(candidate)s", {"candidate" : (self.state.candidate or self.state.inputBuffer) })
```
obilodeau: Mention "heuristic" in the log message and for the json output "candidate" is too broad of a term. Use "candidate_credentials" instead.

pyrdp/mitm/FastPathMITM.py
```diff
@@ -37,6 +37,34 @@ def onClientPDUReceived(self, pdu: FastPathPDU):
...
+
     def onServerPDUReceived(self, pdu: FastPathPDU):
         if self.state.forwardOutput:
-            self.client.sendPDU(pdu)
\ No newline at end of file
+            self.client.sendPDU(pdu)
+
+    def onScanCode(self, scanCode: int, isReleased: bool, isExtended: bool):
```
obilodeau: We should probably store some of the special characters and output them in the logs/json. For example: backspace and tab.

A user has more chances of understanding `P<\b>Passw0rd!` rather than `PPassw0rd!`. I'm not too opinionated about what format it should be but since we are already seeing `\x00` in some fields, we could just pass it like that, so backspace would be `\x08` and tab `\x09`. That's just a suggestion, something fancier would be ok too.

pyrdp/mitm/FastPathMITM.py
```diff
@@ -37,6 +37,34 @@ def onClientPDUReceived(self, pdu: FastPathPDU):
...
+        Handle scan code.
+        """
+        keyName = getKeyName(scanCode, isExtended, self.state.shiftPressed, self.state.capsLockOn)
+
+        if len(keyName) == 1:
+            if not isReleased:
+                self.state.inputBuffer += keyName
+
+        # Left or right shift
+        if scanCode in [0x2A, 0x36]:
```
obilodeau: shouldn't these be constants from somewhere?

pyrdp/mitm/FastPathMITM.py
```diff
@@ -37,6 +37,34 @@ def onClientPDUReceived(self, pdu: FastPathPDU):
...
+            self.state.shiftPressed = not isReleased
+
+        # Caps lock
+        elif scanCode == 0x3A and not isReleased:
```
obilodeau: shouldn't these be constants from somewhere?

pyrdp/mitm/FastPathMITM.py
```diff
@@ -37,6 +37,34 @@ def onClientPDUReceived(self, pdu: FastPathPDU):
...
+            self.state.capsLockOn = not self.state.capsLockOn
+
+        # Return
+        elif scanCode == 0x1C and not isReleased:
```
obilodeau: shouldn't these be constants from somewhere?

pyrdp/mitm/state.py
```diff
@@ -51,6 +51,21 @@ def __init__(self):
         self.forwardOutput = True
         """Whether output from the server should be forwarded to the client"""
 
+        self.loggedIn = False
+        """Keep tracks of the client login status"""
+
+        self.inputBuffer = ""
+        """Used to store what the client types"""
+
+        self.candidate = ""
```
obilodeau: rename candidate_credentials?

obilodeau: or candidateCredentials?

pyrdp/mitm/FastPathMITM.py
Res260: It should, but the only constant we made for that is a dict keycode => friendly name afaik. We should do a constants file from that, we just never did it.

pyrdp/mitm/FastPathMITM.py
Pourliver: Here is an example candidate currently. I tokenized each action the same way to simplify logging / parsing accuracy. 
`MonPass<ctrl-a><\b>Passw0rd!`

Res260:
For reference: slow-path inputs are less used because most clients use fast-path by default for user input, but it happens that they use slow-path only.

Pourliver:
Eh, sorry, I overlooked that part of your first review, I wasn't aware inputs could go through the slow-path. I added that part, and made a base class for both FastPathMITM and SlowPathMITM.

PR #107, opened 2019-05-15
Updated bettercap-rdp-mitm

Pourliver:
Since PyRDP will be presented at the NSEC 2019, I updated the documentation in regard to bettercap's usage with PyRDP.

### DIFF
--- a/docs/bettercap-rdp-mitm.md
+++ b/docs/bettercap-rdp-mitm.md
@@ -1,8 +1,8 @@
 # Run PyRDP alongisde Bettercap
 
 Here is a short tutorial on how to combine the PyRDP and Bettercap to
-redirect all RDP sessions on a LAN segment. By default, Bettercap's arp 
-spoofing module will spoof the entire subnet over a given interface. We 
+redirect all RDP sessions on a LAN segment. By default, Bettercap's arp
+spoofing module will spoof the entire subnet over a given interface. We
 will spawn an instance of PyRDP for each incoming RDP connections if they
 are in our target list, otherwise we will forward them.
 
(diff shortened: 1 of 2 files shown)


docs/bettercap-rdp-mitm.md
```diff
@@ -1,24 +1,60 @@
...
+    cd
+    git clone https://github.com/GoSecure/caplets.git
+    cd caplets
+    sudo make install
+
+## Usage
+
+Start bettercap with :
+
+    sudo bettercap -iface <interface> -caplet <caplet>
```
obilodeau: Using sudo is incompatible with PyRDP installed in a `venv`. Sudo destroys the `PATH` and even `sudo -E` doesn't allow it to be overridden.

@Pourliver had a neat workaround:

```
-eval "set rdp.proxy.command $(which pyrdp-mitm.py);"
```

docs/bettercap-rdp-mitm.md
```diff
@@ -1,24 +1,60 @@
...
 
-Once finished you can remove the iptables rule with
+Basic example :
 
-    iptables -t nat -D PREROUTING -p tcp --dport 3389 -j REDIRECT --to-ports 3389
+    sudo bettercap -iface wlp2s0 -caplet rdp-proxy/rdp-sniffer.cap
```
obilodeau: Same as above `sudo` incompatibility with `venv` note.

docs/bettercap-rdp-mitm.md
```diff
@@ -1,24 +1,59 @@
...
+    cd $GOPATH/src/github.com/GoSecure/
+    git clone https://github.com/GoSecure/bettercap.git -b rdp-mitm
+    cd bettercap
+    go get
+    go build
 
-Have fun!
+Install Caplets by running these commands :
+
+    cd
```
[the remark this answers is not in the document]
Pourliver: FYI, even though the basic caplet works as-is, the other one isn't. They aren't meant to be deleted after installation, since they can be modified where they are git cloned, and the second one needs to be modified, else it won't work.. But I agree that bloating the home directory isn't a great idea.

Pourliver:
@Romounet Since you installed the module recently, did the instructions work out of the box? If so, I'd merge this PR.

Pourliver:
I talked with @Romounet on Slack and updated the documentation to remove ambiguity. I also updated the caplets to better explain the possible configurations. Merging.

PR #109, opened 2019-05-27
Release docs and changelog

obilodeau:
Once accepted, I'll cut the releases in GitHub's UI.

Chose AsciiDoc over markdown because that's already what I maintain with Malboxes and Asciidoctor-reveal.js. Similar release steps and strategies both under SemVer.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -1,0 +1,66 @@
+= {project-name} Changelog
+:project-name: PyRDP
+:uri-repo: https://github.com/GoSecure/pyrdp
+:uri-issue: {uri-repo}/issues/
+
+This document provides a high-level view of the changes introduced in {project-name} by release.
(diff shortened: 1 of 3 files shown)


xshill:
We're gonna have both asciidoc and markdown? :thinking: That seems strange to me.

obilodeau:
I don't see a problem with adding AsciiDoc to Markdown, especially since we are a bigger AsciiDoc shop than markdown and that former format is superior in conciseness.

The project already has reStructuredText docstrings. If having multiple formats would be a real problem shouldn't we have used a reStructuredText README instead of Markdown? So this means multiple formats is ok.

Lastly, as I already stated all my release reflexes (malboxes, asciidoctor-reveal.js) are around AsciiDoc so I'm saving time by using AsciiDoc instead of markdown.

Res260:
Didnt know reStructuredText was meant for readme as well :o

--Émilio

obilodeau:
Releases were cut: https://github.com/GoSecure/pyrdp/releases

Issue #113, opened 2019-06-09
Detect bluekeep scanning/potential exploit attempts


Res260:
Might look into it today

Pourliver:
Merged in #114

PR #114, opened 2019-06-09
Add a warning log message to keep track of bluekeep scanning/exploits attempts

Res260:
@xshill Is it the right place to put this log statement?

Implements #113

### DIFF
--- a/pyrdp/mitm/MCSMITM.py
+++ b/pyrdp/mitm/MCSMITM.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2019 GoSecure Inc.
 # Licensed under the GPLv3 or later.
 #
-
+from logging import LoggerAdapter
 from typing import Callable, Dict
 
 from pyrdp.enum import ClientCapabilityFlag, EncryptionLevel, EncryptionMethod, HighColorDepth, MCSChannelName, \
@@ -27,15 +27,18 @@ class MCSMITM:
     external callback for building MCS channels when a join request is accepted.
     """
 
-    def __init__(self, client: MCSLayer, server: MCSLayer, state: RDPMITMState, recorder: Recorder, buildChannelCallback: Callable[[MCSServerChannel, MCSClientChannel], None]):
+    def __init__(self, client: MCSLayer, server: MCSLayer, state: RDPMITMState, recorder: Recorder,
+                 buildChannelCallback: Callable[[MCSServerChannel, MCSClientChannel], None], log: LoggerAdapter):
         """
         :param client: MCS layer for the client side
         :param server: MCS layer for the server side
         :param state: the RDP MITM shared state
         :param recorder: the recorder for this session
         :param buildChannelCallback: function called when MCS channels are built
+        :param log: logger for the MCS layer.
         """
 
+        self.log = log
         self.client = client
         self.server = server
         self.state = state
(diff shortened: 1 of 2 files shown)


Res260:
I look for clients that open the "MS_T120" virtual channel, which is the first step to trigger bluekeep. This is a pretty good heuristic, and works with this PoC: https://github.com/Ekultek/BlueKeep using the `bluekeep_weaponized_dos.py` file.

PyRDP still crashes, but it's after the log statement, so the statistic still registers.

Note: it does not work with the `bluekeep_poc.py` file, because the client sends an invalid clientData packet (it says that it opens 5 virtual channels, but it sends data about 8 virtual channels, so the packet is invalid, even though it doesnt seem to bother the windows msts server...)

Res260:
Added a commit to fix a bug I introduced.

pyrdp/mitm/MCSMITM.py
```diff
@@ -81,7 +81,8 @@ def onConnectInitial(self, pdu: MCSConnectInitialPDU):
 
         rdpClientDataPDU = rdpClientConnectionParser.parse(gccConferenceCreateRequestPDU.payload)
 
-        if "MS_T120" in map(lambda channelDef: channelDef.name, rdpClientDataPDU.networkData.channelDefinitions):
+        if rdpClientDataPDU.networkData is not None and\
```
Pourliver: This attribute is already validated at line 112. You should move your code there.

Res260: Nice catch, will do.

Res260: done

pyrdp/mitm/MCSMITM.py
```diff
@@ -102,6 +106,9 @@ def onConnectInitial(self, pdu: MCSConnectInitialPDU):
 
         if rdpClientDataPDU.networkData:
             self.state.channelDefinitions = rdpClientDataPDU.networkData.channelDefinitions
+            if "MS_T120" in map(lambda channelDef: channelDef.name, rdpClientDataPDU.networkData.channelDefinitions):
+                self.log.warning("Client tries to open virtual channel 'MS_T120', which most likely means it either"
```
xshill: This log statement is quite a mouthful. The point is to be able to interpret this information (someone's trying to use BlueKeep) very quickly. Wouldn't something like "BlueKeep exploit attempt" be more simple, easier to read and say essentially the same thing?

Also, this seems like INFO level to me. Warning should be for when something went wrong with the actual program or the connection, but it's not something that's very serious. Someone trying to use BlueKeep is information that is actually desirable in a honeypot.

Res260: Yeah I wasnt sure for the warning vs info. I thought warning was good because what follows will probably make pyrdp crash. What do you think?

As for the message, it being long doesnt seem like a problem to me, in fact I think it helps people not knowing details about bluekeep. I dont have strong feelings aboat this however

xshill: It's more about the form than the content. The message is a full english sentence with 22 words. The most important keyword is at the end (BlueKeep), so I have to read a whole 20 words to get the actual meaning. What's actually useful is "BlueKeep attempt" and optionally "channel MS_T120".

I just think the faster you can recognize individual log message types, the better, so they should be concise while still providing the requisite information. This is what I think someone scrolling through a huge text file full of logs would rather have.

Res260:
Fair enough, I'll change it. What about warning vs info?

Pourliver:
I would say this is INFO level, since even though it makes pyrdp crash, its not a protocol error / unexpected packet.

Res260:
Isnt it exactly what it is, an invalid packet?
(What follows)

Issue #104, opened 2019-05-09
Error connection to Windows server 2016 RDS


civic9:
Hi,

I just added to my windows server 2016 Remote Desktop Services Role with two services: Licensing and Session Host.

Now I can't connect through pyrdp. It was all fine before adding RDS Role. Any ideas? .

```
[2019-05-09 16:52:52,439] - INFO - Thomas821467 - pyrdp.mitm.connections.security - Client Info: username = 'administrator\x00', password = '\x
00', domain = '\x00', clientAddress = '....\x00'
[2019-05-09 16:52:52,439] - DEBUG - Thomas821467 - pyrdp.mitm.connections.security - Sending ClientInfoPDU{'payload': b'', 'codePage': 68486165
, 'flags': 213299, 'domain': '\x00', 'username': 'administrator\x00', 'password': '\x00', 'alternateShell': '\x00', 'workingDir': '\x00', 'extr
aInfo': ClientExtraInfo{'payload': b'', 'clientAddressFamily': 2, 'clientAddress': ..., 'clientDir': ...', 'clientTimeZone': b'...', 'clientSessionID': 1, 'performanceFlags': 15, 'autoReconnectCookie': b'', 'dynamicDSTTimeZoneKeyName': b'...', 'dynamicDaylightTimeDisabled': False}}
[2019-05-09 16:52:52,439] - DEBUG - Thomas821467 - pyrdp.mitm.connections.client.mcs - Received MCSPDUType.SEND_DATA_REQUEST
[2019-05-09 16:52:52,445] - DEBUG - Thomas821467 - pyrdp.mitm.connections.server.x224 - Received X224PDUType.X224_TPDU_DATA
[2019-05-09 16:52:52,445] - DEBUG - Thomas821467 - pyrdp.mitm.connections.server.security - Received licensing data
[2019-05-09 16:52:52,445] - DEBUG - Thomas821467 - pyrdp.mitm.connections.server.mcs - Received MCSPDUType.SEND_DATA_INDICATION
[2019-05-09 16:52:52,703] - DEBUG - Thomas821467 - pyrdp.mitm.connections.client.x224 - Received X224PDUType.X224_TPDU_DATA
[2019-05-09 16:52:52,703] - ERROR - GLOBAL - pyrdp - 9 is not a valid SlowPathPDUType
Traceback (most recent call last):
  File "/pyrdp/pyrdp/layer/tcp.py", line 90, in dataReceived
    self.recv(data)
[...]

PR #118, opened 2019-06-12
Fixed licensing issue and added licensing boilerplate

Pourliver:
I have fixed the issue discussed in #104. The actual fix in in SecurityMITM.py, but I have left some boilerplate code inside the (unused) licensing.py if we ever feel like parsing the data.

I had to make a pre-parse validation in order to avoir sending licensing data to the slowpath. I am not quite happy with that method, and I'm open to change it if you have any propositions.

### DIFF
--- a/pyrdp/layer/rdp/security.py
+++ b/pyrdp/layer/rdp/security.py
@@ -135,7 +135,9 @@ class TLSSecurityLayer(SecurityLayer):
         self.securityHeaderExpected = False
 
     def recv(self, data: bytes):
-        if not self.securityHeaderExpected:
+        # Licensing happens in the security layer
+        licensingBytes = b"\x80\x00"
+        if not self.securityHeaderExpected and data[0:2] != b"\x80\x00":
             if self.next is not None:
                 self.next.recv(data)
         else:
--- a/pyrdp/layer/rdp/security.py
+++ b/pyrdp/layer/rdp/security.py
@@ -137,7 +137,7 @@ class TLSSecurityLayer(SecurityLayer):
     def recv(self, data: bytes):
         # Licensing happens in the security layer
         licensingBytes = b"\x80\x00"
-        if not self.securityHeaderExpected and data[0:2] != b"\x80\x00":
+        if not self.securityHeaderExpected and data[0:2] != licensingBytes:
             if self.next is not None:
                 self.next.recv(data)
         else:
--- a/pyrdp/mitm/SecurityMITM.py
+++ b/pyrdp/mitm/SecurityMITM.py
@@ -38,12 +38,13 @@ class SecurityMITM:
         self.recorder = recorder
 
         self.client.createObserver(
+            onLicensingDataReceived = self.onClientLicensingData,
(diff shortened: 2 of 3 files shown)


Res260:
From what you saw of the licensing packets, are there nice information that could be used to fingerprint clients or might be relevant to log in general?

pyrdp/layer/rdp/security.py
```diff
@@ -135,7 +135,9 @@ def __init__(self, parser = BasicSecurityParser()):
         self.securityHeaderExpected = False
 
     def recv(self, data: bytes):
-        if not self.securityHeaderExpected:
+        # Licensing happens in the security layer
+        licensingBytes = b"\x80\x00"
```
Res260: Does that correspond to a flag?

Pourliver: It is, but since this isn't parsed data, it is not in the correct format. Would it be better if I refer to this flag, and parse the 0x80\x00 and compare to the flag?

In pyrdp/enum/rdp.py - line 43
SEC_LICENSE_PKT = 0x0080

Res260: Yeah for that @xshill would have a better idea than me. I have an idea but i’m really not sure if its good

Res260: Reference: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/e13405c5-668b-4716-94b2-1c2654ca1ad4

Pourliver:
The only two interesting things I saw were the "clientUserName" and "clientMachineName", so current username and hostname. With Remmina, the client sent "username" instead of a username to bind the certificate to Remmina instead of a specific user.

I don't think these two have any value since they can be tampered.

Reference : https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpele/c57e4890-9049-421e-9fe8-9a6f9519675a

Res260:
The value might come for fingerprinting purposes, even if they can be tampered. Ex: debbie didnt tamper anything, so he was pretty easy to identify.

Of course its not urgent, simply a nice to have and good to know

Pourliver:
So, after testing, it "works" because the log goes in  the `mitm.json`, but here is what the console output looks like
```
[2019-06-25 12:45:22,225] - INFO - Philip232610 - pyrdp.mitm.connections.tcp - Connection report
```

pyrdp/layer/rdp/security.py
xshill: The licensingBytes you're looking for IS part of a security header. You shouldn't need to check for that if securityHeaderExpected is set properly.

Pourliver: Licensing is an optional part of the protocol, and the boolean "securityHeaderExpected" is false when receiving a licensing PDU. securityHeaderExpected is only used to redirect traffic to the securityLayer when we know we are expecting it in the handshake

xshill: Oh right, since this is TLS it starts at false.

PR #119, opened 2019-06-13
Pyrdp-player improvements

Pourliver:
After using the live pyrdp-player for a few days, with multiple inbound connections at the same time, I had a few improvement ideas :

- Give the new tabs a meaningful name (pyrdp codename or hostname)
- Be clear about which connections are still active, without having to scroll down the log.

The onDisconnection hook wasn't used correctly, so I've removed it, and I added a new rename hook. I rename a tab whenever we receive the clientDataPDU, and on client disconnection. I don't really like using Signals() for that, but it seems common in Qt. Let me know what you think.

### DIFF
--- a/pyrdp/player/LiveEventHandler.py
+++ b/pyrdp/player/LiveEventHandler.py
@@ -7,12 +7,15 @@ from logging import LoggerAdapter
 from pathlib import Path, PosixPath
 from typing import BinaryIO, Dict
 
+from PySide2.QtCore import Signal
 from PySide2.QtWidgets import QTextEdit
 
 from pyrdp.enum import DeviceType, PlayerPDUType
 from pyrdp.layer import PlayerLayer
 from pyrdp.pdu import PlayerDeviceMappingPDU, PlayerDirectoryListingRequestPDU, PlayerDirectoryListingResponsePDU, \
-    PlayerFileDownloadCompletePDU, PlayerFileDownloadRequestPDU, PlayerFileDownloadResponsePDU
+    PlayerFileDownloadCompletePDU, PlayerFileDownloadRequestPDU, PlayerFileDownloadResponsePDU, PlayerPDU
+from pyrdp.parser import ClientConnectionParser
+from pyrdp.player import LiveTab
 from pyrdp.player.FileDownloadDialog import FileDownloadDialog
 from pyrdp.player.filesystem import DirectoryObserver, Drive, File, FileSystem, FileSystemItemType
 from pyrdp.player.PlayerEventHandler import PlayerEventHandler
@@ -25,7 +28,7 @@ class LiveEventHandler(PlayerEventHandler, DirectoryObserver):
     file read events.
     """
 
-    def __init__(self, viewer: QRemoteDesktop, text: QTextEdit, log: LoggerAdapter, fileSystem: FileSystem, layer: PlayerLayer):
+    def __init__(self, viewer: QRemoteDesktop, text: QTextEdit, log: LoggerAdapter, fileSystem: FileSystem, layer: PlayerLayer, connectionRenameTab: Signal, tabInstance: LiveTab):
         super().__init__(viewer, text)
         self.log = log
         self.fileSystem = fileSystem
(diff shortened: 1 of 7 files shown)


Pourliver:
After using the player for a few more days, I fixed some things that were getting annoying.

- We can now see the number of tabs under "Live connections"
- We can now close the tabs with "Ctrl+W", and there will be a prompt if the connection isn't closed. This is configurable with the new "Options" menu, since a user may prefer to let it passthrough to the connection. The Ctrl+W closes the tab by default.
- We can now toggle the "Focus new connection" under the options menu. I was observing a connection earlier, and when I got scanned my tab kept changing, which was really annoying. The auto-focus is on by default.

The current options doesn't persist on close.

pyrdp/player/LiveEventHandler.py
```diff
@@ -25,18 +28,43 @@ class LiveEventHandler(PlayerEventHandler, DirectoryObserver):
...
+        self.handlers[PlayerPDUType.CONNECTION_CLOSE] = self.onConnectionClose
+
+
+    def onClientData(self, pdu: PlayerPDU):
+        """
+        Message the LiveWindow to rename the tab to the hostname of the client
+        """
+
+        clientDataPDU = ClientConnectionParser().parse(pdu.payload)
+        clientName = clientDataPDU.coreData.clientName.strip("\x00")
```
Res260: you might want to use `decodeUTF16LE()` from `helper.py` instead :)

pyrdp/player/LiveWindow.py
```diff
@@ -60,4 +72,25 @@ def sendText(self, text: str):
         tab: LiveTab = self.currentWidget()
 
         if tab is not None:
-            tab.sendText(text)
\ No newline at end of file
+            tab.sendText(text)
+
+    def onTabClosed(self, index):
```
Res260: type hint

pyrdp/player/LiveWindow.py
```diff
@@ -60,4 +72,25 @@ def sendText(self, text: str):
...
+        """
+        Gracefully closes the tab by calling the onClose method
+        :param index: Index of the closed tab
+        """
+        super().onTabClosed(index)
+        self.updateCountSignal.emit()
+
+    def onTabCloseRequest(self, index):
```
Res260: type hint

pyrdp/player/MainWindow.py
```diff
@@ -91,4 +117,17 @@ def sendText(self):
...
             return
 
-        self.liveWindow.sendText(text)
\ No newline at end of file
+        self.liveWindow.sendText(text)
+
+    def toggleFocusNewTab(self):
+        self.options["focusNewTab"] = not self.options.get("focusNewTab")
+
+    def togglecloseTabOnCtrlW(self):
```
Res260: typo for camelCase

pyrdp/player/LiveEventHandler.py
Pourliver: Parsing this string as UTF-16 gives garbage. Plus, I used that code from somewhere else in the code base.  This is already a parsed PDU, it has already been decoded as UTF-16LE

Res260: Ahh, I didnt remember it could give trailing null bytes!

pyrdp/player/LiveTab.py
```diff
@@ -23,7 +23,7 @@ class LiveTab(BaseTab, DirectoryObserver):
     Tab playing a live RDP connection as data is being received over the network.
     """
 
-    connectionClosed = Signal(object)
+    connectionRenameTab = Signal(object, str, bool)
```
xshill: Why have rename and closed in the same signal? If those are 2 separate events (it seems like they are), then there should be 2 signals.

pyrdp/player/LiveWindow.py
```diff
@@ -37,15 +42,22 @@ def onConnection(self) -> asyncio.Protocol:
...
+
+        self.updateCountSignal.emit()
         self.queue.put(tab)
 
-    def onConnectionClosed(self, tab: LiveTab):
+    def renameLivePlayerTab(self, tab: LiveTab, name: str, closed: bool):
         index = self.indexOf(tab)
-        text = self.tabText(index)
-        self.setTabText(index, text + " - Closed")
+        if closed:
```
xshill: Proof that they are separate: you check if one argument is present, otherwise you check the other argument.

pyrdp/player/LiveTab.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, parent: QWidget = None):
         self.layers = layers
         self.rdpWidget = rdpWidget
         self.fileSystem = FileSystem()
-        self.eventHandler = LiveEventHandler(self.widget, self.text, self.log, self.fileSystem, self.layers.player)
+        self.eventHandler = LiveEventHandler(self.widget, self.text, self.log, self.fileSystem, self.layers.player, self.connectionRenameTab, self)
```
xshill: It seems a bit strange to pass signals around to me. In C++ signals are just methods and not objects, so this wouldn't be possible to do trivially. Consider moving the signals to the live event handler directly and exposing the signals through other signals or just doing something like `self.connectionRenameTab = self.eventHandler.connectionRenameTab`.

It seems like you can even connect signals to other signals just like in C++ so this should not involve too much duplication: https://groups.google.com/d/msg/pyside/xlRLJfGmGYM/3uKYkf98jckJ

It makes the class more self-contained and easier to understand IMO.

Pourliver: PyQt signals needs to be defined on a class of type QObject, and the LiveEventHandler isn't one which is why I did it this way. I'll look into it

pyrdp/player/LiveTab.py
Pourliver: Fixed in latest commit.

pyrdp/player/LiveTab.py
Pourliver: Done.

PR #117, opened 2019-06-12
Add a statcounter that prints statistics about the connection

Res260:
Logs various data about the connection when the client disconnects. Looks like this:
image
each field is separated so it can be used as filters and to build visualizations in ES/Kibana.

What I implemented for now:

```python
    CONNECTION_TIME = "connectionTime"
    # Duration (in secs) for the TCP connection

    CLIENT_SERVER_RATIO = "clientServerRatio"
    # Ratio of the # of messages coming from the client vs from the server. High value (>1) means high client interaction.

    TOTAL_INPUT = "totalInput"
    # # of messages coming from the client to the server.

    TOTAL_OUTPUT = "totalOutput"
    # # of messages coming from the server to the client.

    IO_INPUT = "input"
    # Packet coming from the client to the server for the io channel

    IO_INPUT_FASTPATH = "fastPathInput"
    # Packet coming from the client to the server for the io channel as a fastpath packet

    IO_INPUT_SLOWPATH = "slowPathInput"
    # Packet coming from the client to the server for the io channel as a slowpath packet

    IO_OUTPUT = "output"
    # Packet coming from the server to the client for the io channel

    IO_OUTPUT_FASTPATH = "fastPathOutput"
    # Packet coming from the server to the client for the io channel as a fastpath packet

    IO_OUTPUT_SLOWPATH = "slowPathOutput"
    # Packet coming from the server to the client for the io channel as a slowpath packet

    MCS = "mcs"
    # Packet Coming from either end for any channel

    MCS_OUTPUT = "mcsOutput"
    # Packet Coming from the server to the client for any channel

    MCS_OUTPUT_ = "mcsOutput_"
    # Packet Coming from the server to the client for a given channel (must append channel # after it)

    MCS_INPUT = "mcsInput"
    # Packet Coming from the client to the server for any channel

    MCS_INPUT_ = "mcsInput_"
    # Packet Coming from the client to the server for a given channel (must append channel # after it)

    VIRTUAL_CHANNEL = "virtualChannel"
    # Packet Coming from either end for any virtual channel that doesnt have a specific implementation (ex clipboard)

    VIRTUAL_CHANNEL_INPUT = "virtualChannelInput"
    # Packet Coming from the client to the server for any virtual channel that doesnt have a specific implementation (ex clipboard)

    VIRTUAL_CHANNEL_OUTPUT = "virtualChannelOutput"
    # Packet Coming from the server to the client for any virtual channel that doesnt have a specific implementation (ex clipboard)

[...]

PR #123, opened 2019-06-19
Added client's IP and clientData to allow client identification

### LINKED ISSUE #122: The IP of the connection isn't logged if exited before receiving ClientInfoPDU
Currently, most of the incoming connections are scanners. Scanners won't send a ClientInfoPDU since it's later in the protocol.

With our current logging, we have no way of identifying WHO connected to pyrdp. This would be really useful to correlate data. Should this be added in #117, or appended to 
`New Client connected`?

Ex:
`New client connected from 127.0.0.1`

Pourliver:
Fixes #122 . I also log clientData, which normally is the hostname, but tools use it to identify "what" did the connection, just like an user-agent.

### DIFF
--- a/pyrdp/layer/tcp.py
+++ b/pyrdp/layer/tcp.py
@@ -57,7 +57,14 @@ class TwistedTCPLayer(IntermediateLayer, Protocol):
         When the TCP handshake is completed, notify the observer.
         """
         self.connectedEvent.set()
-        self.observer.onConnection()
+
+        try:
+            # Connection to client
+            ip = self.transport.client[0]
+            self.observer.onConnection(ip)
+        except AttributeError:
+            # Connection to server / attacker
+            self.observer.onConnection()
 
     def connectionLost(self, reason=connectionDone):
         """
--- a/pyrdp/layer/tcp.py
+++ b/pyrdp/layer/tcp.py
@@ -57,14 +57,7 @@ class TwistedTCPLayer(IntermediateLayer, Protocol):
         When the TCP handshake is completed, notify the observer.
         """
         self.connectedEvent.set()
-
-        try:
-            # Connection to client
-            ip = self.transport.client[0]
-            self.observer.onConnection(ip)
-        except AttributeError:
-            # Connection to server / attacker
-            self.observer.onConnection()
+        self.observer.onConnection()
 
     def connectionLost(self, reason=connectionDone):
         """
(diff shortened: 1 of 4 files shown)


pyrdp/layer/tcp.py
```diff
@@ -57,7 +57,14 @@ def connectionMade(self):
         When the TCP handshake is completed, notify the observer.
         """
         self.connectedEvent.set()
-        self.observer.onConnection()
+
+        try:
+            # Connection to client
+            ip = self.transport.client[0]
+            self.observer.onConnection(ip)
+        except AttributeError:
```
Res260: There must be a cleaner way to do that 🤔

pyrdp/mitm/TCPMITM.py
```diff
@@ -62,12 +62,12 @@ def detach(self):
         self.client.removeObserver(self.clientObserver)
         self.server.removeObserver(self.serverObserver)
 
-    def onClientConnection(self):
+    def onClientConnection(self, ip):
```
Res260: Type hint

pyrdp/mitm/TCPMITM.py
```diff
@@ -62,12 +62,12 @@ def detach(self):
...
         """
         Log the fact that a new client has connected.
         """
         self.connectionTime = time.time()
-        self.log.info("New client connected")
+        self.log.info("New client connected from %(client_ip)s", {"client_ip": ip})
```
Res260: Use camelCase

pyrdp/mitm/mitm.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, log: SessionLogger, config: MITMConfig):
         self.x224 = X224MITM(self.client.x224, self.server.x224, self.getLog("x224"), self.state, serverConnector, self.startTLS)
         """X224 MITM component"""
 
-        self.mcs = MCSMITM(self.client.mcs, self.server.mcs, self.state, self.recorder, self.buildChannel)
+        self.mcs = MCSMITM(self.client.mcs, self.server.mcs, self.state, self.recorder, self.buildChannel, self.getLog("mcs"))
```
Res260: Nothing to change but it might cause a conflict with #117 so we just need to watch out for that

pyrdp/layer/tcp.py
Pourliver: I haven't seen anything defined at this moment (or before) that allowed me to be 100% sure that the current connection is a client. Still, I'll have another look.

Pourliver: I overlooked a small detail, much simpler and cleaner now. Thanks! :slightly_smiling_face:

Issue #127, opened 2019-07-03
Windows isn't supported 


Pourliver:
From what I've seen, there are two things which block PyRDP from running on Windows.

1. Notify2 uses dbus, which doesn't exist on Windows, as reffered in #101. This is fixed in my branch [windows_support](https://github.com/GoSecure/pyrdp/tree/windows_support)
2. pycrypto doesn't work out of the box on Windows. It require some manual modification to pycrypto itself. Pycrypto isn't supported anymore, the replacement is pycryptodome.

We could (and probably should) replace pycrypto by pycryptodome. If anyone has tried to use PyRDP on Windows, I'd like to know if there's anything I'm missing.

Res260:
Note that PyCrypto is used for the RC4 encryption (when TLS is not used), so keep that in mind when testing!

obilodeau:
If it's only used for RC4 and nothing else (no HMAC, etc.) then we should probably inline an open version. It's simple

Res260:
@xshill would be the reference for that.

Pourliver:
From my test, the RC4 encryption works the same in pycrypto and pycryptodome. The only part that changed was the RSA module. So handling RC4 isn't a problem. But I will still be testing the changes over RDP and TLS security :)

Res260:
Implemented in #129

PR #128, opened 2019-07-04
Replaced pycrypto with pycryptodome

Pourliver:
We had many reason to replace pycrypto with pycryptodome. First, it hasn't been supported in 4-5 years, and second it blocks #127. 

The only "downside" with pycryptodome is that we cannot use raw RSA to encrypt and decrypt. Since it is considered unsafe to encrypt/decrypt without padding, they removed the ability to do so, but it is needed for the "RDP" security level. I took the logic behind the original pycrypto RSA module, and made a new class out of it with the required functions.

I've tested most of the different use case, and everything works as intended :tada:


pyrdp/security/crypto.py
```diff
@@ -12,6 +15,39 @@
 Cryptographic utility functions
 """
 
+# Adapted from https://github.com/dlitz/pycrypto/blob/master/lib/Crypto/PublicKey/_slowmath.py
+# Not meant to be a safe implementation. We only need raw RSA without padding.
```
Res260: Should go in the class' docstring

pyrdp/security/crypto.py
```diff
@@ -12,6 +15,39 @@
...
+class RSA:
+    """
+    Class for encrypting or decrypting data with RSA
+    """
+
+    def __init__(self, key: RsaKey):
+        self.key = key
+
+    def encrypt(self, plaintext: bytes):
```
Res260: return type

pyrdp/security/crypto.py
```diff
@@ -12,6 +15,39 @@
...
+        m = bytes_to_long(plaintext)
+        ciphertext = pow(m, self.key.e, self.key.n)
+        return long_to_bytes(ciphertext)
+
+    def decrypt(self, ciphertext: bytes):
```
Res260: return type

Pourliver:
Done! Should be good now @xshill

PR #129, opened 2019-07-04
Added Windows support

Pourliver:
As said in #127, I removed Notify2 and dbus references when using Windows. I also moved "Installing on Windows" up a notch in the README, since that we can now actually install it without any hacks.

I have tested on a brand new environment without any problems. Note that this PR won't work on Windows without #128, but since they are really distinct I think they should be kept apart.


bin/pyrdp-player.py
```diff
@@ -42,12 +42,15 @@ def prepareLoggers(logLevel: int, outDir: Path):
...
-    uiLogger.addHandler(notifyHandler)
+    # https://docs.python.org/3/library/os.html
+    if os.name != "nt":
+        notifyHandler = NotifyHandler()
+        notifyHandler.setFormatter(notificationFormatter)
 
+        uiLogger = logging.getLogger(LOGGER_NAMES.PLAYER_UI)
+        uiLogger.addHandler(notifyHandler)
+    else:
+        print("Notifications are not supported for your platform, they will be disabled.")
```
Res260: use `pyRDPLogger.warning()`

README.md
```diff
@@ -52,7 +52,7 @@ PyRDP was [first introduced in a blogpost](https://www.gosecure.net/blog/2018/12
 ## Supported Systems
 PyRDP should work on Python 3.6 and up.
 
-This tool has been tested to work on Python 3.6 on Linux (Ubuntu 18.04). It has not been tested on OSX and Windows.
+This tool has been tested to work on Python 3.6 on Linux (Ubuntu 18.04). It has not been tested on OSX.
```
xshill: Do you mean "tested to work on Linux [...] and Windows"?

Pourliver: Yup :)

Pourliver:
@xshill Sorry for the delay, went all-in on the crawler lol

Issue #125, opened 2019-06-20
Crash when scanned by some bluekeep scanners


Pourliver:
PyRDP currently crashes when someone tries to "bluekeep" it. For example, the following crash comes from [rdpscan](https://github.com/robertdavidgraham/rdpscan). I have tested a few other scanners (I recommend [BKScan](https://github.com/nccgroup/BKScan)) and they don't produce the same crash. Maybe we should investigate if the crash is on our side, or if the scanners have some bad hardcoded PDUs.

```
[2019-06-20 15:28:14,938] - ERROR - GLOBAL - pyrdp - Exception occurred when receiving: 0300019a02f0807f6582018e0401010401010101ff30200202002202020002020200000202000102020000020200010202ffff020200023020020200010202000102020001020200010202000002020001020204200202000230200202ffff0202fc170202ffff0202000102020000020200010202ffff020200020482011b000500147c00018112000800100001c00044756361810401c0d800040008002003580201ca03aa09040000280a00007200640070007300630061006e0000000000000000000000000000000000000004000000000000000c0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001ca010000000000ff00070001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000100000004c00c00090000000000000002c00c00000000000000000003c01400010000004d535f543132300080800000

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/twisted/python/log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/usr/local/lib/python3.6/dist-packages/twisted/python/log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/usr/local/lib/python3.6/dist-packages/twisted/python/context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/local/lib/python3.6/dist-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/usr/local/lib/python3.6/dist-packages/twisted/internet/asyncioreactor.py", line 136, in _readOrWrite
    why = method()
  File "/usr/local/lib/python3.6/dist-packages/twisted/internet/tcp.py", line 243, in doRead
    return self._dataReceived(data)
[...]

PR #136, opened 2019-07-16
Added logger to TwisterTCPLayer for error handling and made PyRDP more "scanner compliant"

Pourliver:
Errors are now correctly tagged with the correct sessionID instead of global.

I also removed the enum HighColorDepth from the coreDataPDU since most scanners and scripts use a non-valid value. It serves no real purpose since the server respond to a bad HighColorDepth by enforcing another one.


Pourliver:
Also fixes #125

PR #140, opened 2019-07-29
Added recursive download and tab icon when a filesystem is shared

Pourliver:
We can now right click and download a folder recursively. I made the downloadDialog non-modal, since a modal is a blocking action, meaning that we cannot focus the original window. I did the same thing to the download completion popup.

I also added a jobs queue that allows us to queue up downloads, they starts as soon as the previous download has ended. Practical when we are downloading a huge zip / folder, and we want to keep looking for interesting file.

The logic may be confusing, because of the nature of the observers, I had to make sure that only one job was executing at a time.

The 3 queues have distinct jobs. `jobsQueue` holds the action requested by the user. When a job is being processed, the sub-jobs goes into `directoryDownloadQueue` and `fileDownloadQueue`. When we download a folder, we start by downloading all files inside the folder, then we do the same for each sub-folder.

I've worked on this for a while so I don't know what I should explain design wise lol. Let me know if anything is unclear.


pyrdp/player/LiveEventHandler.py
```diff
@@ -80,6 +90,46 @@ def onListDirectory(self, deviceID: int, path: str):
...
+    def addToDownloadQueue(self, item: Union[File, Directory], targetPath: str, dialog: FileDownloadDialog):
+        job = (item, targetPath, dialog)
+
+        self.jobsQueue.add(job)
+
+        if self.currentDownload == None:
+            self.dispatchDownload()
+
+    def dispatchDownload(self):
+        # We download each file of a directory before enumerating other directories.
```
Res260: docstring ` ``` ` instead of comment `#`

pyrdp/player/LiveEventHandler.py
```diff
@@ -80,6 +90,46 @@ def onListDirectory(self, deviceID: int, path: str):
...
+        # Otherwise we may fill the stack with too many queued items.
+        # When we're done with both file and directories, we start an other queued job
+
+        # Request download of a queued file
+        if len(self.fileDownloadQueue) != 0:
```
Res260: `myQueue.empty()` exists I think

Res260: (see elifs as well)

pyrdp/player/LiveEventHandler.py
```diff
@@ -139,7 +217,26 @@ def onFileDownloadRequested(self, file: File, targetPath: str, dialog: FileDownl
         self.downloadFiles[remotePath] = targetFile
         self.downloadDialogs[remotePath] = dialog
 
-        pdu = PlayerFileDownloadRequestPDU(self.layer.getCurrentTimeStamp(), parent.deviceID, file.getFullPath())
+        pdu = PlayerFileDownloadRequestPDU(self.layer.getCurrentTimeStamp(), parent.deviceID, remotePath)
+        self.layer.sendPDU(pdu)
+
+    def onDirectoryDownloadRequested(self, directory: Directory, targetPath: str, dialog: FileDownloadDialog):
```
Res260: since this method got way bigger docstring would help

pyrdp/player/LiveEventHandler.py
Pourliver: It doesn't, in python `if not myQueue` is falsy if the queue is empty. Even though its very pythonic, it's not very verbose. I think using `len(myQueue) != 0` should stay, as seen everywhere else in the code base

pyrdp/player/LiveEventHandler.py
Pourliver: Will do! I'll add more documentation to the flow of the module in general.

pyrdp/player/LiveEventHandler.py
Res260: im confused: https://docs.python.org/3/library/queue.html#queue.Queue.empty

Pourliver: Thats the definition of the queue class, which is not what I'm using. The Queue class is meant for multi-threaded application.

Res260: Damn, I overlooked that! I understand.

Res260:
Merged master into the branch, updated changelog and tested. Ready to merge.

Issue #137, opened 2019-07-19
Improvement to client drive explorer / filesystem mapping


Pourliver:
After playing with the drive explorer, and sharing a big folder, it was really painful to look around for files.

I think we should be able to recursively download a folder. I had the idea when I saw a .git folder, but it could apply to so much more.

I also thing we should be able to configure some kind of auto-scrapping. We would need to have some kind of crawler that could look around for some interesting extensions like txt, xls, doc, and all that should be configurable. Since it is very rare to have someone connect, and the odds of that connection happening while we are monitoring with the PyRDP player is very low, we should have a way to automatically download files that the crawler found without the pyrdp-player.

Maybe that the crawler / auto-download should be built in the pyrdp-mitm directly, and the recursive download in the pyrdp-player. What do you think?

Pourliver:
We would also need a way to rapidly identify connections with a mounted drive in the pyrdp-player. I think we should add an icon to the connection tab, and even prompt before closing the connection.

xshill:
Auto-download in pyrdp-mitm makes sense to avoid missing something if the player is not open. Recursive downloads could also go in pyrdp-mitm for the same reason. You could use it to automatically download all .git folders for example.

Res260:
Implemented in #140 and #141

PR #141, opened 2019-08-01
Added shared drive crawler

Pourliver:
I added the automatic shared drive crawler I had in mind, as shared a little in #137.

This has been done after #140, so its been easier to implement since I knew the download flow and quirks.

I had to create a crawler_config folder since I didn't like the idea of mixing classes definition and plaintext files.

I tried to put more documentation in this class since the core logic may be hard to follow. I've tested it a lot and it works amazingly well, but please do test it.


Res260:
Idea: log in a file each file, even if not downloaded

pyrdp/mitm/FileCrawler.py
```diff
@@ -0,0 +1,277 @@
...
+        localPath = f"{basePath}{remotePath}"
+
+        self.log.info("Saving %(remotePath)s to %(localPath)s", {"remotePath": remotePath, "localPath": localPath})
+
+        try:
+            # Create parent directory, don't raise error if it already exists
+            Path(localPath).parent.mkdir(parents=True, exist_ok=True)
+            targetFile = open(localPath, "wb")
+        except Exception as e:
+            self.log.error("Cannot save file: %(exception)s", {"exception": str(e)})
```
Res260: `log.exception` would be better

pyrdp/mitm/FileCrawler.py
```diff
@@ -0,0 +1,277 @@
...
+            return
+
+        self.downloadFiles[remotePath] = targetFile
+        self.deviceRedirection.sendForgedFileRead(file.deviceID, remotePath)
+
+    def listDirectory(self, deviceID: int, path: str, download: bool = False):
```
Res260: docstring would help

pyrdp/mitm/config.py
```diff
@@ -53,6 +53,15 @@ def __init__(self):
         self.payloadDuration: int = None
         """Amount of time the payload should take to complete, in milliseconds"""
 
+        self.disableCrawler: bool = False
+        """Whether the crawler should be crawling or not"""
```
Res260: should be disabled or not*, to match the boolean value

pyrdp/mitm/crawler_config/match.txt
```diff
@@ -0,0 +1,25 @@
...
+*/*.csv
+*/*.xls*
+*/*.doc*
+*/*.ppt*
+
+# Possible malware samples
+*/*.bat
+*/*.exe
+*/*.dll
+*/*.ps1
```
Res260: .py? possibly other languages as well

bin/pyrdp-mitm.py
```diff
@@ -164,6 +164,9 @@ def main():
     parser.add_argument("--payload-powershell-file", help="PowerShell script to run automatically upon connection (as -EncodedCommand)", default=None)
     parser.add_argument("--payload-delay", help="Time to wait after a new connection before sending the payload, in milliseconds", default=None)
     parser.add_argument("--payload-duration", help="Amount of time for which input / output should be dropped, in milliseconds. This can be used to hide the payload screen.", default=None)
+    parser.add_argument("--disable-crawler", help="Disable automatic shared drive scraping", action="store_true")
+    parser.add_argument("--crawler-match-file", help="File to be used by the crawler to chose what to download when scraping the client shared drives.")
+    parser.add_argument("--crawler-ignore-file", help="File to be used by the crawler to chose what folders to avoid when scraping the client shared drives.")
```
Res260: put default value?

pyrdp/mitm/crawler_config/match.txt
Pourliver: Do you have any input on this? @obilodeau This file is what we currently download from the disk, folder AND files.

Pourliver:
> Idea: log in a file each file, even if not downloaded

To add to this idea, since we talked about it in person, the goal is to keep track of what the crawler is seeing, so we can further improve the default match / ignore regexes later.

pyrdp/mitm/crawler_config/match.txt
Pourliver: I think that crawling code files would give us a lot of thrash and would expand the crawl time. I went with compiled files and scripts that work out-of-the-box on Windows / Linux.

bin/pyrdp-mitm.py
Pourliver: Since the configuration takes places in `bin/pyrdp-mitm.py` and the default file is in `pyrdp/mitm/crawler_config/match.txt`, I don't think moving the default value in the configuration parser is a good idea. Don't forget that the `bin/` content gets moved to the user binaries folder during installation, and we lose direct path references.

pyrdp/mitm/FileCrawler.py
```diff
@@ -0,0 +1,277 @@
...
+        :param filePath: Unix-style path of the file.
+        :param isDirectory: True if the file is a directory.
+        """
+
+        self.deviceID = deviceID
+        self.name = name
+        self.path = filePath
+        self.isDirectory = isDirectory
+
+class FileCrawler(DeviceRedirectionMITMObserver):
```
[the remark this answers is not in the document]
Pourliver: I'm not too sure the value that it brings, but I can rename it

bin/pyrdp-mitm.py
Res260: makes sense

pyrdp/mitm/FileCrawler.py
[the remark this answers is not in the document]
Res260: It was for consistence and verboseness, but I dont think we're constant in the rest of the code base so doesnt matter for now

Pourliver:
The PR is ready for another review. I added a new logger that is not from the root logger, so we could log in another file without messing with STDOUT. This new file keeps track of every file we crawled.

Res260:
Oops commented on the commet instead of the PR:

Besides the exception() log TODO, LGTM!

Pourliver:
Sorry, but I'm not too sure I understand whats wrong @Res260 , could you please elaborate a bit?

Res260:
I think I saw you added a log.exception() somewhere else in the code, so I
thought it would be easy to implement instead of leaving a TODO in the code.

Pourliver:
You're right, I totally missed this one, my bad! Fixed in latest commit.

Res260:
Lgtm

Pourliver:
Fixed a small problem with the parser where the last line of each file got truncated, and simplified the match file since `*/*.doc*` can also match `*/*.docker`, which is unintended.

bin/pyrdp-mitm.py
```diff
@@ -169,6 +191,9 @@ def main():
     parser.add_argument("--payload-powershell-file", help="PowerShell script to run automatically upon connection (as -EncodedCommand)", default=None)
     parser.add_argument("--payload-delay", help="Time to wait after a new connection before sending the payload, in milliseconds", default=None)
     parser.add_argument("--payload-duration", help="Amount of time for which input / output should be dropped, in milliseconds. This can be used to hide the payload screen.", default=None)
+    parser.add_argument("--disable-crawler", help="Disable automatic shared drive scraping", action="store_true")
```
xshill: I think this is such a big feature that it should be enabled consciously, so I'll change this to opt-in.

PR #155, opened 2019-10-02
Added missing package dep

robeving:
This adds a missing package required to run the player in docker


xshill:
6 other commits have been pushed. Did you mean to only merge the first commit?

PR #183, opened 2020-02-06
player: Remove trailing null bytes when printing text (fixes #182)

### LINKED ISSUE #182: Stacktrace in the player: embedded null character
```
$ pyrdp-player.py out/rdp_replay_XXX.pyrdp 
[2020-02-05 16:34:35,815] - INFO - pyrdp.player - Listening for connections on 127.0.0.1:3000
Traceback (most recent call last):
  File "/home/olivier/Documents/gosecure/src/pyrdp/pyrdp/ui/qt.py", line 150, in runOnMainThread
    target()
  File "/home/olivier/Documents/gosecure/src/pyrdp/pyrdp/player/PlayerEventHandler.py", line 60, in <lambda>
    self.viewer.mainThreadHook.emit(lambda: self.onPDUReceived(pdu, True))
  File "/home/olivier/Documents/gosecure/src/pyrdp/pyrdp/player/PlayerEventHandler.py", line 66, in onPDUReceived
    self.handlers[pdu.header](pdu)
  File "/home/olivier/Documents/gosecure/src/pyrdp/pyrdp/player/PlayerEventHandler.py", line 111, in onClipboardData
    self.writeText(f"CLIPBOARD DATA: {clipboardData}")
  File "/home/olivier/Documents/gosecure/src/pyrdp/pyrdp/player/PlayerEventHandler.py", line 52, in writeText
    self.text.insertPlainText(text)
ValueError: embedded null character
```

obilodeau:


Res260:
If I'm not mistaken, copying an ascii string in Windows (which would most likely be copied as UTF-16) could make the parsing fail (since its little endian, ABC would be encoded as 42 00 43 00 44 00)

obilodeau:
The object text is of type `str` which is encoding independent. According to [python docs](https://docs.python.org/3.8/library/stdtypes.html#textseq):

> Strings are immutable sequences of Unicode code points.

So here I don't think we can expect wide strings. To make sure I tested in ipython:

```
In [3]: s = str(b"\x42\x00\x43\x00\x44\x00", encoding="utf-16")                                                                                                                                                                                                                                                               

In [4]: s.rstrip("\x00")                                                                                                                                                                                                                                                                                                                     
Out[4]: 'BCD'

In [5]: s = str(b"\x42\x00\x43\x00\x44\x00\x00\x00", encoding="utf-16")                                                                                                                                                                                                                                                       

In [6]: s                                                                                                                                                                                                                                                                                                                     
Out[6]: 'BCD\x00'

In [7]: s.rstrip("\x00")                                                                                                                                                                                                                                                                                                      
Out[7]: 'BCD'
```

So I think the fix is correct. If text was a byte string you would have been right.

PR #186, opened 2020-03-02
Add a CLI argument to disable active clipboard stealing

Res260:
uses only passive stealing instead

- Add pyrdp-generated files to .gitignore

Tested using mstsc.exe on win 10 to a win7 target machine

### DIFF
--- a/bin/pyrdp-mitm.py
+++ b/bin/pyrdp-mitm.py
@@ -1,20 +1,21 @@
 #!/usr/bin/env python3
 
+import argparse
 #
 # This file is part of the PyRDP project.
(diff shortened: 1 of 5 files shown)


bin/pyrdp-mitm.py
```diff
@@ -1,20 +1,21 @@
 #!/usr/bin/env python3
 
+import argparse
```
alxbl: Should go below license?

Res260: fixed

Res260:
should be good now @alxbl

Res260:
nvm

Res260:
ok now it should be good lol

PR #187, opened 2020-03-03
Add support for Graphics Device Interface Acceleration Extensions (MS-RDPEGDI)

### LINKED ISSUE #50: Implement GDI+Drawing Orders
Recent versions of RDP use GDI+ for graphic rendering (although PyRDP MITM disables most of it, it might not be possible in future RDP versions). This library is written by Microsoft, but a Unix implementation exists: https://www.mono-project.com/docs/gui/libgdiplus/

The plan would be to do a wrapper for this library in Python to call the needed methods in the PyRDP Player.

### LINKED ISSUE #162: The connection speed of pyrdp-mitm.py is very slow How to improve the connection speed? Rdpy is much faster than your pyrdp
The connection speed of pyrdp-mitm.py is very slow How to improve the connection speed? Rdpy is much faster than your pyrdp

alxbl:
This pull request adds support for the MS-RDPEGDI extension and drawing orders.

I'm opening it now despite being only roughly halfway done with the changes in order to give some visibility as to the overall changes so far.

**What's done**
- [X] Parsing for the message types defined in MS-RDPEGDI.
- [X] Pipeline to get those messages to the Player
- [X] API for GDI frontends

**What's left to do**
- [x] Support for Capability Propagation in the Drawing Order Parser
- [x] Support for GlyphV2 parsing in the Drawing Order Parser
- [x] Qt5 Frontend for drawing orders
- [x] Glyph Rendering
- [x] Blending/Composition Mode Interop between RDP/GDI and Qt
- [x] Cleanup pass
- [X] Documentation
- [x] Testing
- ~[ ] Implement NINEGRID support and rendering (currently downgrading)~

Closes #50
Fixes #162

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+# ignore = 
+max-line-length = 160
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
 # ignore = 
-max-line-length = 160
+max-line-length = 120
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -29,6 +29,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * Added `--disable-active-clipboard` switch to prevent clipboard request injection
 * Added `--no-downgrade` switch to prevent protocol downgrading where possible {uri-issue}189[#189]
 * Added `--no-files` switch to prevent extracting transferred files {uri-issue}195[#195]
+* Added `--gdi` switch to enable drawing orders ({uri-issue}50[#50])
 
 === Bug fixes
 
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -29,7 +29,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * Added `--disable-active-clipboard` switch to prevent clipboard request injection
 * Added `--no-downgrade` switch to prevent protocol downgrading where possible {uri-issue}189[#189]
 * Added `--no-files` switch to prevent extracting transferred files {uri-issue}195[#195]
-* Added `--gdi` switch to enable drawing orders ({uri-issue}50[#50])
+* Added `--gdi` MITM switch to enable accelerated graphics pipeline (MS-RDPEGDI) ({uri-issue}50[#50])
 
 === Bug fixes
 
--- a/README.md
+++ b/README.md
@@ -391,8 +391,7 @@ remove the `ip rule`.
 
 Tells the MITM to allow clients to use [Graphics Device Interface Acceleration][gdi] Extensions to stream
 drawing orders instead of raw bitmaps. The advantage of this mode is a significant reduction in required bandwidth
-for high resolution connections. PyRDP player support has been recently added and it is possible that some issues
-exist. As a result, this is currently an opt-in setting.
+for high resolution connections.
 
 [gdi]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegdi/745f2eee-d110-464c-8aca-06fc1814f6ad
 
--- a/README.md
+++ b/README.md
@@ -393,6 +393,9 @@ Tells the MITM to allow clients to use [Graphics Device Interface Acceleration][
 drawing orders instead of raw bitmaps. The advantage of this mode is a significant reduction in required bandwidth
 for high resolution connections.
 
+Note that some GDI drawing orders are currently unimplemented because they appear to be unused.
+If you have a replay which contains any unsupported or untested order, do not hesitate to share it with the project maintainers so that support can be added as required. (Make sure that the trace does not contain sensitive information)
+
 [gdi]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegdi/745f2eee-d110-464c-8aca-06fc1814f6ad
 
 ### Using the PyRDP Player
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ PyRDP was [introduced in 2018](https://www.gosecure.net/blog/2018/12/19/rdp-man-
     + [Other MITM arguments](#other-mitm-arguments)
       - [--no-downgrade](#--no-downgrade)
       - [--transparent](#--transparent)
+      - [`--gdi`: Accelerated Graphics Pipeline](#--gdi-accelerated-graphics-pipeline)
   * [Using the PyRDP Player](#using-the-pyrdp-player)
     + [Playing a replay file](#playing-a-replay-file)
     + [Listening for live connections](#listening-for-live-connections)
@@ -386,6 +387,14 @@ This setup is a base example and might be much more complex depending on your en
 To cleanup, you should delete the created routes, remove the custom routing table from `rt_tables` and
 remove the `ip rule`.
 
+##### `--gdi`: Accelerated Graphics Pipeline
+
+Tells the MITM to allow clients to use [Graphics Device Interface Acceleration][gdi] Extensions to stream
+drawing orders instead of raw bitmaps. The advantage of this mode is a significant reduction in required bandwidth
+for high resolution connections. PyRDP player support has been recently added and it is possible that some issues
+exist. As a result, this is currently an opt-in setting.
+
+[gdi]: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegdi/745f2eee-d110-464c-8aca-06fc1814f6ad
 
 ### Using the PyRDP Player
 Use `pyrdp-player.py` to run the player.
--- a/pyrdp/enum/orders.py
+++ b/pyrdp/enum/orders.py
@@ -1,3 +1,9 @@
+#
+# This file is part of the PyRDP project.
+# Copyright (C) 2019 GoSecure Inc.
+# Licensed under the GPLv3 or later.
+#
+
 """
 Enumerations for Drawing Orders.
 """
--- a/pyrdp/enum/orders.py
+++ b/pyrdp/enum/orders.py
@@ -1,0 +1,72 @@
+"""
+Enumerations for Drawing Orders.
+"""
+
+from enum import IntEnum
+
+
+class DrawingOrderControlFlags(IntEnum):
+    """
+    https://msdn.microsoft.com/en-us/library/cc241574.aspx
+    """
+    TS_STANDARD = 0x01
+    TS_SECONDARY = 0x02
+    TS_BOUNDS = 0x04
+    TS_TYPE_CHANGE = 0x08
+    TS_DELTA_COORDS = 0x10
+    TS_ZERO_BOUNDS_DELTAS = 0x20
+    TS_ZERO_FIELD_BYTE_BIT0 = 0x40
+    TS_ZERO_FIELD_BYTE_BIT1 = 0x80
+
+
+class Primary(IntEnum):
+    DSTBLT = 0x00
+    PATBLT = 0x01
+    SCRBLT = 0x02
+    DRAW_NINE_GRID = 0x07
+    MULTI_DRAW_NINE_GRID = 0x08
+    LINE_TO = 0x09
+    OPAQUE_RECT = 0x0A
+    SAVE_BITMAP = 0x0B
+    MEMBLT = 0x0D
+    MEM3BLT = 0x0E
+    MULTI_DSTBLT = 0x0F
+    MULTI_PATBLT = 0x10
+    MULTI_SCRBLT = 0x11
+    MULTI_OPAQUE_RECT = 0x12
+    FAST_INDEX = 0x13
+    POLYGON_SC = 0x14
+    POLYGON_CB = 0x15
+    POLYLINE = 0x16
+    FAST_GLYPH = 0x18
+    ELLIPSE_SC = 0x19
+    ELLIPSE_CB = 0x1A
+    GLYPH_INDEX = 0x1B
+
+
+class Secondary(IntEnum):
+    BITMAP_UNCOMPRESSED = 0x00
+    CACHE_COLOR_TABLE = 0x01
+    CACHE_BITMAP_COMPRESSED = 0x02
+    CACHE_GLYPH = 0x03
+    BITMAP_UNCOMPRESSED_V2 = 0x04
+    BITMAP_COMPRESSED_V2 = 0x05
+    CACHE_BRUSH = 0x07
+    BITMAP_COMPRESSED_V3 = 0x08
+
+
+class Alternate(IntEnum):
+    SWITCH_SURFACE = 0x00
+    CREATE_OFFSCREEN_BITMAP = 0x01
+    STREAM_BITMAP_FIRST = 0x02
+    STREAM_BITMAP_NEXT = 0x03
+    CREATE_NINE_GRID_BITMAP = 0x04
+    GDIPLUS_FIRST = 0x05
+    GDIPLUS_NEXT = 0x06
+    GDIPLUS_END = 0x07
+    GDIPLUS_CACHE_FIRST = 0x08
+    GDIPLUS_CACHE_NEXT = 0x09
+    GDIPLUS_CACHE_END = 0x0A
+    WINDOW = 0x0B
+    COMPDESK_FIRST = 0x0C
+    FRAME_MARKER = 0x0D
--- a/pyrdp/enum/orders.py
+++ b/pyrdp/enum/orders.py
@@ -1,6 +1,6 @@
 #
 # This file is part of the PyRDP project.
-# Copyright (C) 2019 GoSecure Inc.
+# Copyright (C) 2020 GoSecure Inc.
 # Licensed under the GPLv3 or later.
 #
 
--- a/pyrdp/parser/rdp/fastpath.py
+++ b/pyrdp/parser/rdp/fastpath.py
@@ -17,7 +17,6 @@ from pyrdp.security import RC4Crypter, RC4CrypterProxy
 
 from pyrdp.parser.parser import Parser
 from pyrdp.parser.rdp.bitmap import BitmapParser
-from pyrdp.parser.rdp.orders import OrdersParser
 from pyrdp.parser.rdp.security import BasicSecurityParser
 
 from pyrdp.logging import log
--- a/pyrdp/parser/rdp/fastpath.py
+++ b/pyrdp/parser/rdp/fastpath.py
@@ -8,16 +8,19 @@ from binascii import hexlify
 from io import BytesIO
 
 from pyrdp.core import Uint16BE, Uint16LE, Uint8
-from pyrdp.enum import DrawingOrderControlFlags, EncryptionMethod, FastPathInputType, \
+from pyrdp.enum import EncryptionMethod, FastPathInputType, \
     FastPathOutputCompressionType, FastPathOutputType, FastPathSecurityFlags, FIPSVersion, ParserMode
-from pyrdp.logging import log
-from pyrdp.parser.parser import Parser
-from pyrdp.parser.rdp.bitmap import BitmapParser
-from pyrdp.parser.rdp.security import BasicSecurityParser
 from pyrdp.pdu import FastPathBitmapEvent, FastPathEventRaw, FastPathMouseEvent, FastPathOrdersEvent, FastPathPDU, \
-    FastPathScanCodeEvent, SecondaryDrawingOrder
+    FastPathScanCodeEvent
 from pyrdp.pdu.rdp.fastpath import FastPathEvent, FastPathOutputEvent, FastPathUnicodeEvent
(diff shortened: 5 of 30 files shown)


pyrdp/enum/orders.py
```diff
@@ -0,0 +1,72 @@
+"""
```
Res260: dont forget the pyrdp header

pyrdp/parser/rdp/fastpath.py
```diff
@@ -409,33 +400,19 @@ def writeBitmapEvent(self, stream: BytesIO, event: FastPathBitmapEvent):
...
-        orderCount = Uint16LE.unpack(stream)
-        orderData = stream.read(size - 2)
-
-        assert len(orderData) == size - 2
+        """
+        Parse the order events from a TS_FP_UPDATE_ORDERS.
+        This is specified in MS-RDPEGDI.
+        """
+        payload = stream.read(size)
+        assert len(payload) == size
```
Res260: I don't remember if we use `assert` in PyRDP, but putting an error message as a second argument would help for the end user to understand exactly what went wrong

pyrdp/enum/orders.py
alxbl: I'm going to do a final pass to clean up the code and make sure all of that stuff is present once the code is done.

pyrdp/parser/rdp/fastpath.py
alxbl: Asserts are not meant to be user-consumable errors. If this condition is false, the stream is either malformed or corrupted. I will be adding error handling code that skips unknown PDUs and optionally logs them to serve as unit tests.

alxbl: I've added the code I was talking about in db459c9

pyrdp/enum/orders.py
alxbl: Fixed in e6f8bf5

alxbl:
The branch is now in a state where I feel comfortable getting a thorough code review.

I'm in the process of doing more exhaustive tests using Windows 7 and Windows Server 2012.

README.md
```diff
@@ -331,6 +332,15 @@ to be established. The following are currently not affected by this switch and w
 RDP traffic using Wireshark and keep the TLS master secrets. Whenever PyRDP adds support for additional extensions, 
 it would then become possible to extract a valid RDP replay file from the raw network capture.
 
+##### `--gdi`: Accelerated Graphics Pipeline
+
+Tells the MITM to allow clients to use [Graphics Device Interface Acceleration][gdi] Extensions to stream
+drawing orders instead of raw bitmaps. The advantage of this mode is a significant reduction in required bandwidth
+for high resolution connections. PyRDP player support has been recently added and it is possible that some issues
```
Res260: maybe write the date? Feel like a year from now it'll be confusing

alxbl: You're right, I'll actually rephrase this so that it doesn't even mention "recently added". That's redundant.

pyrdp/player/gdi/__init__.py
```diff
@@ -0,0 +1,7 @@
+#
+# This file is part of the PyRDP project.
+# Copyright (C) 2019 GoSecure Inc.
```
Res260: 2020?

Res260: Its 2019 everywhere i think

pyrdp/parser/rdp/slowpath.py
```diff
@@ -625,3 +627,40 @@ def writePointerCapability(self, capability: PointerCapability, stream: BytesIO)
...
+        stream.read(3)  # Padding
+
+        keys = stream.read(64 * (num0 + num1 + num2 + num3 + num4))
+        return PersistentCacheKeysPDU(header, num0, num1, num2, num3, num4,
+                                      total0, total1, total2, total3, total4, keys, bBitMask)
+
+    def writePersistentCacheKeys(self, s: BytesIO, pdu: PersistentCacheKeysPDU):
+        # Only send the first PDU with an empty list and drop the rest.
+        # TODO: Find a way to cleanly drop the entire packet instead.
+        Uint16LE.pack(0, s)
```
Res260: `s.write(b"\x00")` works too

pyrdp/player/gdi/draw.py
```diff
@@ -0,0 +1,702 @@
...
+        t = (color >> 5) & 0x3F
+        t = (t << 2) + t // 4 // 2
+        g = min(t, 255)
+
+        t = (color >> 11) & 0x1F
+        t = (t << 3) + t // 4
+        r = min(t, 255)
+
+    elif bpp == 15:
+        pass
```
Res260: maybe raise instead to avoid potential future bugs?

pyrdp/parser/rdp/orders/parse.py
```diff
@@ -0,0 +1,355 @@
...
+
+from .frontend import GdiFrontend
+
+from .secondary import CacheBitmapV1, CacheColorTable, CacheGlyph, CacheBitmapV2, CacheBrush, CacheBitmapV3
+from .alternate import CreateOffscreenBitmap, SwitchSurface, CreateNineGridBitmap, \
+    StreamBitmapFirst, StreamBitmapNext, GdiPlusFirst, GdiPlusNext, GdiPlusEnd, GdiPlusCacheFirst, \
+    GdiPlusCacheNext, GdiPlusCacheEnd, FrameMarker
+from .primary import PrimaryContext as Context
+
+LOG = logging.getLogger('pyrdp.fastpath.rdp.orders')
```
Res260: there are constants for this

pyrdp/parser/rdp/orders/frontend.py
```diff
@@ -0,0 +1,182 @@
...
+class GdiFrontend:
+    """
+    Interface for objects that implement GDI.
+
+    This class provides abstract methods to be used by modules
+    interested in listening and acting upon context updates.
+    Its primary purpose is for the PyRDP player to render the
+    remote desktop.
+
+    NOTE: Unimplemented methods will act as No-Op.
```
Res260: shouldnt they raise or at least log?

pyrdp/player/gdi/__init__.py
obilodeau: You can't bump copyright years everywhere. Proper copyright handling means bumping when you modify the file. So as we change files, we need to update the year. Top-level `LICENSE` file can be updated as soon as one file in the package is.

Res260: i think its a new file so it should be 2020, right?

pyrdp/player/gdi/draw.py
alxbl: `QColor.fromRgb()` will raise on undefined symbols already

pyrdp/parser/rdp/orders/frontend.py
alxbl: Not for the base class. The current frontend implementation `GdiFrontend` warns on the first unsupported order it sees.

pyrdp/parser/rdp/orders/parse.py
alxbl: What is the point of having the logger names be constants in an external module? The logger namespace is local to the module by definition, and is not supposed to be used by other modules if the point is to namespace log output. Do we have plans to rename loggers in the future? Do we have plans to re-use the same loggers in multiple areas?

This is a lot of unnecessary fragmentation in my opinion. The furthest I'm willing to concede is to have the top level logger name be a constant.

pyrdp/player/gdi/__init__.py
alxbl: Yes, it should be 2020. I'll go through the files that are in the PR and make sure it's 2020 everywhere.

pyrdp/parser/rdp/orders/parse.py
Res260: It depends on the logger use case. In this case, its for a "program status" kind of use, so it's fine using module name as logger names. What I meant is that you should use `__name__` instead of `pyrdp.fastpath.rdp.orders`. In this module `__name__` will be equal to `pyrdp.parser.rdp.orders` which give pretty much the same information as in your hardcoded string. It allows easier refactor, as changing names of modules should not require to look in string values in a program (or at least not unnecessarily)

"constants" was probably a poor choice of word, I think the correct term is "magic variable"

alxbl: I had not thought of using `__name__`. That is indeed better than a hardcoded string.

alxbl:
Rebased this on master since there have been a lot of changes.

obilodeau:
@alxbl: new conflicts. sorry :disappointed:

alxbl:
Conflicts are resolved. I want this to land before mp4 conversion if possible.

pyrdp/player/gdi/draw.py
```diff
@@ -0,0 +1,702 @@
...
+
+    Note that `CacheBitmapV3` is only used in 32 bits-per-pixel which
+    PyRDP currently does not support due to RDP6.0 compression not
+    being implemented yet.
+
+    If you have a replay which contains any unsupported or untested
+    order, do not hesitate to share it with the project maintainers so
+    that support can be added as required.
+
+    (Make sure that the trace does not contain any sensitive information)
```
obilodeau, line 103: This should be added to the README as well. It is too important to be only here.

pyrdp/player/gdi/draw.py
```diff
@@ -0,0 +1,702 @@
...
+        p.setBrushOrigin(b.x, b.y)
+
+    # Drawing API.
+    def onBounds(self, bounds: Bounds):
+        """Called on bounding rectangle updates."""
+        self.bounds = bounds
+
+    def dstBlt(self, state: DstBlt):
+        """Destination-only blitting operation."""
+        LOG.debug(state)
```
obilodeau, line 180: can we afford these calls? I know that logging in tight loops should be avoided because the interpreter can't compile out the `LOG.debug()` call from being there and then later realizes it is not useful.

Last time I had to do something like this we would do something like `if DEBUG: <logging statement>` to avoid the more expensive function call and the interpreter could compile out the statement if DEBUG was false. It was in Perl so this would need to be reverified for python.

That said, we shouldn't do this until profiling prooves us that this is expensive.

pyrdp/player/gdi/draw.py
```diff
@@ -0,0 +1,702 @@
...
+                self.glyphs.addFragment(cid, fid, fragment)
+            else:
+                glyph = self.glyphs.get(cid, instr)
+                i = ctx.offset(i, data)
+                ctx.draw(glyph, p)
+        self._end(p)
+
+    def _unimplemented(self):
+        if not self._warned:
+            LOG.warning('One or more unimplemented drawing orders called! Expect lossy rendering.')
```
obilodeau, line 572: :+1: 

Curious about how many hits we will have in the wild :smile:

pyrdp/player/gdi/raster.py
```diff
@@ -0,0 +1,390 @@
...
+        elif c == 'S':
+            stack.append(src)
+        elif c == 'P':
+            if pal is None:
+                raise SyntaxError('Palette is not present.')
+            stack.append(pal)
+        else:
+            lhs = stack.pop()
+            rhs = None
+            res = lhs  # TODO: Actually perform the operation.
```
obilodeau, line 372: what does this TODO task mean? This isn't implemented?

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
+# ignore = 
+max-line-length = 160
```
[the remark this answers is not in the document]
obilodeau: We had that discussion on the side yesterday (after I wrote this). We can leave it as is.

pyrdp/parser/rdp/orders/alternate.py
```diff
@@ -0,0 +1,232 @@
...
+    @staticmethod
+    def parse(s: BytesIO) -> 'CreateOffscreenBitmap':
+        self = CreateOffscreenBitmap()
+
+        self.flags = Uint16LE.unpack(s)
+        self.id = self.flags & 0x7FFF
+        self.cx = Uint16LE.unpack(s)
+        self.cy = Uint16LE.unpack(s)
+
+        if self.flags & 0x8000 != 0:
```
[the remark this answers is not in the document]
obilodeau: you can ignore this one, it's a mask on the first bit on an 32-bit int, pretty obvious

.flake8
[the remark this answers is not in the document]
Res260: if it helps
Shill and I agreed on 120 at the beginning of the project, so most of the codebase should fit with 120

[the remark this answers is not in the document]
alxbl: I think 120 is a more sensible value. I remember setting 160 because I was having a lot of annoying lint warnings all over the place.

CHANGELOG.adoc
```diff
@@ -29,6 +29,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * Added `--disable-active-clipboard` switch to prevent clipboard request injection
 * Added `--no-downgrade` switch to prevent protocol downgrading where possible {uri-issue}189[#189]
 * Added `--no-files` switch to prevent extracting transferred files {uri-issue}195[#195]
+* Added `--gdi` MITM switch to enable accelerated graphics pipeline (MS-RDPEGDI) ({uri-issue}50[#50])
```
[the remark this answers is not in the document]
alxbl: I do think GDI should be supported by default, but I did want to gate it at first since I'm sure there are unidentified bugs. If you believe we should go straight into GDI (and fix bugs as they come), then sure, I can flip it around.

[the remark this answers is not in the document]
obilodeau: How about this: we merge this as-is, we open a ticket to consider flipping that behavior and we milestone that ticket with `vNext`.

So we will benefit from this code `$now`, we deploy master in our usual spot :wink: and our users can do to. Before our next release, we evaluate whether this is ready enough to be flipped or not.

What do you think?

pyrdp/parser/rdp/fastpath.py
```diff
@@ -368,27 +361,24 @@ def isCompressed(self, header: int) -> bool:
...
+        compressionFlags = Uint8.unpack(stream) if self.isCompressed(header) else None
         size = Uint16LE.unpack(stream)
 
-        eventType = header & 0xf
-        fragmentation = header & 0b00110000 != 0
-
-        if fragmentation:
-            log.debug("Fragmentation is present in output fastpath event packets."
-                      " Not parsing it and saving to FastPathOutputUpdateEvent.")
-            return FastPathOutputEvent(header, compressionFlags, payload=stream.read(size))
```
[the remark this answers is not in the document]
alxbl: Nothing, the function will parse the first fragment of the order, which will create the proper type in Python, which is used down the line where the actual reassembly of fragments happens (In the Player code). It actually simplifies logic further down the pipeline since we don't have to re-check and re-parse the FastPathOutputType as a specific type a second time.

CHANGELOG.adoc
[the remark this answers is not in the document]
Res260: I'd argue that since the MITM does not parse GDI, there is very little chance that this PR breaks the MITM and that any bug can be fixed with minor version releases, so it should be the default

pyrdp/parser/rdp/fastpath.py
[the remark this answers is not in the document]
obilodeau: Awesome. Resolving.

pyrdp/parser/rdp/orders/parse.py
```diff
@@ -0,0 +1,355 @@
...
+# Parser Lookup Tables
+_pri = [
+    OrdersParser._parse_dstblt,                # 0x00
+    OrdersParser._parse_patblt,                # 0x01
+    OrdersParser._parse_scrblt,                # 0x02
+    None,                                      # 0x03
+    None,                                      # 0x04
+    None,                                      # 0x05
+    None,                                      # 0x06
+    OrdersParser._parse_draw_nine_grid,        # 0x07
```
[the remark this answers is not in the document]
alxbl: It isn't, pure coincidence :) ... well, I guess all drawing orders are always specified in the same order, so the cap index happens to match the order ID.

pyrdp/parser/rdp/orders/parse.py
```diff
@@ -0,0 +1,355 @@
...
+    OrdersParser._parse_polygon_cb,            # 0x15
+    OrdersParser._parse_polyLine,              # 0x16
+    None,                                      # 0x17
+    OrdersParser._parse_fast_glyph,            # 0x18
+    OrdersParser._parse_ellipse_sc,            # 0x19
+    OrdersParser._parse_ellipse_cb,            # 0x1A
+    OrdersParser._parse_glyph_index,           # 0x1B
+]
+
+assert len(_pri) == 0x1C
```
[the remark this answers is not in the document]
alxbl: It was initially to ensure that I didn't miss any. It just stayed in because I forgot about it.

pyrdp/parser/rdp/orders/primary.py
```diff
@@ -0,0 +1,1231 @@
...
+        if self.ctx.field(8):
+            self.nYSrc = read_coord(s, self.ctx.deltaCoords, self.nYSrc)
+        if self.ctx.field(9):
+            self.bg = read_rgb(s)
+        if self.ctx.field(10):
+            self.fg = read_rgb(s)
+
+        self.brush.update(s, self.ctx.fieldFlags >> 10)
+
+        if self.ctx.field(16):
```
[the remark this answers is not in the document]
alxbl: An enum for what? the field numbers? I don't think they are ever going to change order, and it would add one enum per message, and not really make the code more readable. It would just add like 300-400+ LoCs.

I did contemplate having some kind of loop, or defining the field handlers in an array, but it felt overengineered for no reason and even confusing. Macros would actually have been nice to have here.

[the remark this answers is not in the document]
alxbl: In case it's not clear what this code is doing, each field in the message is optional, and the message contains a bitfield that indicates the presence of field N with its Nth bit. Of course that bitfield is also compressed using the 2 MSB to indicate how many bytes of the bitfield are set to all 0.

This check has to be done for every field, for every message.

pyrdp/player/gdi/cache.py
```diff
@@ -0,0 +1,202 @@
...
+GDI Cache Management Layer.
+"""
+
+from PySide2.QtCore import QSize
+from PySide2.QtGui import QBrush, QImage, QBitmap
+
+from pyrdp.parser.rdp.orders.common import Glyph
+
+
+class BitmapCache:
```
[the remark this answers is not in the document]
alxbl: The player still works in headless mode. These files are only imported when starting in full GUI mode, which will not work if PySide2 is not there. This is the same behavior as the bitmap rendering.

pyrdp/player/gdi/cache.py
```diff
@@ -0,0 +1,202 @@
...
+        Add an entry to the cache.
+
+        :returns: True if the entry is a fresh entry, False if it replaced an existing one.
+        """
+        if cid not in self.caches:
+            self.caches[cid] = {}
+        cache = self.caches[cid]
+        cache[idx] = entry
+
+    def evict(self, cid: int, idx: int) -> bool:
```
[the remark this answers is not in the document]
alxbl: There is actually a maximum number of cache entries which get recycled. I used a dictionary in Python because it's a lot easier (and cleaner) than managing bounded arrays. Eviction is requested by the server, but once we drop a pane, all of this code should get garbage collected. A single connection is bound by how much cache entries it can have, so there is an upper limit to the memory consumption.

This should not be a problem.

pyrdp/player/gdi/draw.py
alxbl: The `str()` representation is not evaluated if the verbosity level is not active, so these are just NOOP function calls. I did not have any trouble rendering captures with `INFO` level verbosity, and `DEBUG` was mostly bottlenecked by stderr flushing, so I don't think this is an issue. It will be trivial to fix if we ever get performance issue reports

pyrdp/player/gdi/draw.py
alxbl: :D

pyrdp/player/gdi/raster.py
alxbl: Yes. It ended up not being necessary.

alxbl:
All the things have been flakified.

CHANGELOG.adoc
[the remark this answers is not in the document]
alxbl: I think @obilodeau has the right approach. We're still a while off from releasing, so we can land this and revisit making GDI default when we're ready for 1.0.

I also agree that any bugs that come up are going to be purely aesthetic/rendering bugs, since recording is unaffacted. The problem is I don't want to become a blocker for people with video that  needs to be viewed.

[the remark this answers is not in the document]
obilodeau: Good. Opened #209.

Issue #170, opened 2019-11-28
Convert pyrdp recordings to mp4/mkv


OXERY:
Hi there,
did anybody manage to export the pyrdp-recordings as video files?

Res260:
It has not been done, as far as I know. We used a screen recorder when we needed it. It would be a nice addition to PyRDP!

alxbl:
Here are my thoughts on this:

We could repurpose the pyrdp-pcap-converter tool to be just `pyrdp-convert` which would support all our format conversions:

- pcap to replay
- replay to mp4
- pcap to mp4 (via replay)

I'm planning to use PyAV since it appears to provide most of the bindings that we need in libavcodec.
We would eventually depend on libavcodec anyway if we want to support the other video pipelines. For now the dependency would be optional when pyrdp-convert is used.

This is the next significant feature I intend to work on unless something else comes up.

Cheers,
Alex

PR #188, opened 2020-03-10
Add PyRDP Pcap Extractor

obilodeau:
This has been in a branch aside for a while now since it was not very polished. However, it is valuable so I think we should aim to introduce it to master soon.

TODO
- [ ] move Warning from README to the tool itself
- [ ] rename the tool. Suggestion: `pyrdp-pcap-extractor`
- [ ] feature: extract transferred files (or files dragged and dropped)
- [ ] test with a pcap in GDI passthrough mode and make sure that we can extract all interesting virtual channels (keyboard, files, etc.)


Res260:
doesnt `pyrdp-pcap-extractor` suggest that PyRDP can somehow extract pcaps? ``pyrdp-replay-extractor` seems more suitable

alxbl:
I'm thinking this could be merged with the mp4 conversion issue (#170) and renamed to `pyrdp-convert` to be the global tool for converting. Thoughts?

alxbl:
I would vote to close this in favor of #199 which contains the refactored `pyrdp-convert` as discussed. Let me know.

alxbl:
Closing as superseded by #199

PR #189, opened 2020-03-13
feat: Added `--no-downgrade` mode.

alxbl:
This is a rework of the `--gdi-passthrough` branch that covers a broader set of passthrough options.

After some internal discussion, a mode that avoids fingerprinting by not downgrading any of the connection features makes more sense than per-extension support switches, so I renamed the switch to `--no-downgrade` and added some documentation to explain its current state, what it can and cannot do, and its impacts on the PyRDP player.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,7 +17,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * Default Docker Compose command now `pyrdp-mitm -h` to avoid confusing crash on `docker-compose up` ({uri-issue}173[#173])
 * Documentation updates and fixes ({uri-issue}165[#165], {uri-issue}166[#166], {uri-issue}172[#172])
 * Added `--disable-active-clipboard` switch to prevent clipboard request injection
-* Added `--no-downgrade` switch to prevent protocol downgrading where possible.
+* Added `--no-downgrade` switch to prevent protocol downgrading where possible {uri-issue}189[#189]
 
 
 === Bug fixes
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,6 +17,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * Default Docker Compose command now `pyrdp-mitm -h` to avoid confusing crash on `docker-compose up` ({uri-issue}173[#173])
 * Documentation updates and fixes ({uri-issue}165[#165], {uri-issue}166[#166], {uri-issue}172[#172])
 * Added `--disable-active-clipboard` switch to prevent clipboard request injection
+* Added `--no-downgrade` switch to prevent protocol downgrading where possible.
 
 
 === Bug fixes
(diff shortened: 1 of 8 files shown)


pyrdp/mitm/SlowPathMITM.py
```diff
@@ -59,21 +59,22 @@ def onConfirmActive(self, pdu: ConfirmActivePDU):
...
-        pdu.parsedCapabilitySets[CapabilityType.CAPSTYPE_ORDER].orderSupport = b"\x00" * 32
-
         # Disable virtual channel compression
         if CapabilityType.CAPSTYPE_VIRTUALCHANNEL in pdu.parsedCapabilitySets:
             pdu.parsedCapabilitySets[CapabilityType.CAPSTYPE_VIRTUALCHANNEL].flags = VirtualChannelCompressionFlag.VCCAPS_NO_COMPR
 
-        # Override the bitmap cache capability set with null values.
-        if CapabilityType.CAPSTYPE_BITMAPCACHE in pdu.parsedCapabilitySets:
-            pdu.parsedCapabilitySets[CapabilityType.CAPSTYPE_BITMAPCACHE] = Capability(CapabilityType.CAPSTYPE_BITMAPCACHE, b"\x00" * 36)
+        if not self.state.config.downgrade:
```
alxbl: Oops, this should be `if self.state.config.downgrade:` :S

CHANGELOG.adoc
```diff
@@ -17,6 +17,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 * Default Docker Compose command now `pyrdp-mitm -h` to avoid confusing crash on `docker-compose up` ({uri-issue}173[#173])
 * Documentation updates and fixes ({uri-issue}165[#165], {uri-issue}166[#166], {uri-issue}172[#172])
 * Added `--disable-active-clipboard` switch to prevent clipboard request injection
+* Added `--no-downgrade` switch to prevent protocol downgrading where possible.
```
obilodeau:
```suggestion
* Added `--no-downgrade` switch to prevent protocol downgrading where possible {uri-issue}189[#189]
```
nitpick: Only use dots if there's more than one sentence. Add Link to issue or PR.

README.md
```diff
@@ -306,6 +306,30 @@ After 5 seconds, input / output is restored back to normal.
 #### Other MITM arguments
 Run `pyrdp-mitm.py --help` for a full list of arguments.
 
+##### `--no-downgrade`
```
obilodeau: Because of this new section the ToC will need to be regenerated. They used an online tool to do so. It kind of sucks that this is a manual step to be honest. Migrating this file to AsciiDoc would eliminate that problem. That said, just updating the ToC would be fine for now.

README.md
```diff
@@ -306,6 +306,30 @@ After 5 seconds, input / output is restored back to normal.
...
+
+- FIPS Encryption
+- Non-TLS encryption protocols
+- ClientInfo compression
+- Virtual Channel compression
+
+**NOTE**: If being able to eventually replay the full session is important, a good solution is to record the raw 
+RDP traffic using Wireshark and keep the TLS master secrets. Whenever PyRDP adds support for additional extensions, 
+it would then become possible to extract a valid RDP replay file from the raw network capture.
+
```
obilodeau: Awesome doc!

PR #190, opened 2020-03-13
Multiple Player improvements

### LINKED ISSUE #151: Make the GUI an optional feature?
Due to #150 issue (QT 5 no longer provides binaries for x86 32-bit) PyRDP doesn't work on 32-bit OS.

However, it would be possible to specify the GUI as an optional feature using Python's `extra_requires` section in `setup.py`: https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies. Also, we could specify which console entrypoint requires the optional dependency (spoiler, it's `pyrdp-player.py`).

This would make PyRDP's `pyrdp-mitm.py` work in headless environments and 32-bit environments in one go. We could also get rid of the DBus dependency and its related issues.

If we implement it, we should remember to:
* get rid of dbus-python and notify2 alongside with PySide2
* adjust README to pull in optional dependency by default but explain the alternative
* revert 0794b5e04

Instead of doing the work I opened an issue because I would like to know if people think the headless option is valuable. I won't do it if no one is to use it.

### LINKED ISSUE #163: pyrdp-player.py without a GUI but with logs
As mentioned in #151 [here](https://github.com/GoSecure/pyrdp/issues/151#issuecomment-537431429):

> I'd also like to run pyrdp-player.py without a GUI so that it just produces a log file result

I think this is a good idea that should be documented.

alxbl:
- [X] Optional GUI dependencies (`pip install -U -e '.[GUI]'`)
- [X] Auto-play single replay files
- [x] Headless `pyrdp-player` (`--headless`)

Closes #151, Closes #163

### DIFF
--- a/bin/pyrdp-player.py
+++ b/bin/pyrdp-player.py
@@ -19,12 +19,16 @@ import logging.handlers
 import sys
 import os
 
-from PySide2.QtWidgets import QApplication
+try:
+    from PySide2.QtWidgets import QApplication
+    from pyrdp.player import MainWindow
+    HAS_GUI = True
+except ModuleNotFoundError:
+    HAS_GUI = False
 
 from pyrdp.logging import LOGGER_NAMES, NotifyHandler
-from pyrdp.player import MainWindow
 
-def prepareLoggers(logLevel: int, outDir: Path):
+def prepareLoggers(logLevel: int, outDir: Path, headless: bool):
     logDir = outDir / "logs"
     logDir.mkdir(exist_ok = True)
 
@@ -42,15 +46,20 @@ def prepareLoggers(logLevel: int, outDir: Path):
     pyrdpLogger.addHandler(fileHandler)
     pyrdpLogger.setLevel(logLevel)
 
-    # https://docs.python.org/3/library/os.html
-    if os.name != "nt":
-        notifyHandler = NotifyHandler()
-        notifyHandler.setFormatter(notificationFormatter)
+    if not headless and HAS_GUI:
+        # https://docs.python.org/3/library/os.html
+        if os.name != "nt":
+            try:
+                notifyHandler = NotifyHandler()
+                notifyHandler.setFormatter(notificationFormatter)
 
-        uiLogger = logging.getLogger(LOGGER_NAMES.PLAYER_UI)
-        uiLogger.addHandler(notifyHandler)
-    else:
-        pyrdpLogger.warning("Notifications are not supported for your platform, they will be disabled.")
+                uiLogger = logging.getLogger(LOGGER_NAMES.PLAYER_UI)
+                uiLogger.addHandler(notifyHandler)
+            except Exception:
+                # No notification daemon or DBus, can't use notifications.
+                pass
+        else:
+            pyrdpLogger.warning("Notifications are not supported for your platform, they will be disabled.")
 
 def main():
     """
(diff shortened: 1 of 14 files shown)


bin/pyrdp-player.py
```diff
@@ -63,19 +72,35 @@ def main():
...
     logLevel = getattr(logging, args.log_level)
-    prepareLoggers(logLevel, outDir)
+    prepareLoggers(logLevel, outDir, args.headless)
 
-    app = QApplication(sys.argv)
-    mainWindow = MainWindow(args.bind, int(args.port), args.replay)
-    mainWindow.show()
+    if not HAS_GUI and not args.headless:
+        logging.error('Headless mode is not specified and PySide2 is not installed. Install PySide2 to use the graphical user interface.')
+        exit(127)
```
obilodeau: The `exit` method exists in this context? Shouldn't this be `sys.exit()`?

pyrdp/player/HeadlessEventHandler.py
```diff
@@ -0,0 +1,154 @@
+#
+# This file is part of the PyRDP project.
+# Copyright (C) 2019 GoSecure Inc.
```
obilodeau:
```suggestion
# Copyright (C) 2020 GoSecure Inc.
```

bin/pyrdp-player.py
alxbl: It does, but I think it's an interpreter built-in, so it might not exist on all implementations of Python. I'll change it to `sys.exit`

alxbl:
> Question: I think the existing replay event handler didn't log file operations between the client and the server or mouse clicks (w/ coordinates) am I correct?

It doesn't. The mouse is rendered visually and it's pretty difficult/spammy to render the coordinates nicely in text only. We could try to log mouse clicks.

For the file events, we can add support in both places whenever we add it.

I'll make the modifications and merge this.

Cheers,

alxbl:
Oh lord. I just realized that the entire `keyboard.py` file which processes scancodes depends on Qt. I had reinstalled PySide2 so it didn't error, but my last tests are failing. I'm going to have to do a bit more work on this.

obilodeau:
> Oh lord. I just realized that the entire `keyboard.py` file which processes scancodes depends on Qt. I had reinstalled PySide2 so it didn't error, but my last tests are failing. I'm going to have to do a bit more work on this.

Great catch :+1: 

I was under the impression that we borrowed scan codes from FreeRDP but it could not be the case. Maybe if not using QT we could re-import them from FreeRDP. I let you investigate all the options.

obilodeau:
> It doesn't. The mouse is rendered visually and it's pretty difficult/spammy to render the coordinates nicely in text only. We could try to log mouse clicks.

Yes, I was thinking just on clicks you log the X, Y at that moment. Might seem not to add a lot of value but the way I see it is if I replay capture and have no event, I would be wondering if the user is doing everything through mouse whereas if I see the clicks then I would know for sure. This is out of scope for this PR btw.

> 
> For the file events, we can add support in both places whenever we add it.

:+1:

alxbl:
> I was under the impression that we borrowed scan codes from FreeRDP but it could not be the case. Maybe if not using QT we could re-import them from FreeRDP. I let you investigate all the options.

We have the scan code mappings to `Qt.Key` enums, I just need to make them not rely on the Qt import, so it should just be a matter of a few editor macros and minor code changes to use the new mappings.

alxbl:
Tested without PySide2 installed end-to-end to make sure I didn't miss anything. I fixed a few things in the process and added mouse click reporting in the replays (both headless and GUI).

I'll give you some time to look my latest commit over before merging.

I can see a few opportunities for refactoring (like having a better event handler hierarchy: `Headless -> Player -> Live`) and some code de-duplication, but I feel like this is clean enough right now.

robeving:
I'm excited to test this. I'm currently processing PCAPs of thousands of connections through some hacked up changed to pyrdp that use a framebuffered X server.

alxbl:
@robeving That sounds like a very creative workaround! Let me know how headless mode goes for you. Feel free to open issues if you run into any problems when trying it.

robeving:
It works and about 1000x faster!

PR #191, opened 2020-03-17
feat: Add support for logging configuration files.

alxbl:
This exposes all logging configuration (and eventually command line flags) through a configuration file.

The idea is to give users more control on what is output and eventually simplify launching the tool.
The PR is a draft because I want to clean up the existing logging code and add support for the player before merging.

The plan is to have a `mitm.ini` and `player.ini` which contains all configuration parameters for PyRDP. We could also theoretically have a single `config.ini` which all tools read, but that would increase log configuration complexity.

### DIFF
--- a/bin/pyrdp-player.py
+++ b/bin/pyrdp-player.py
@@ -6,15 +6,16 @@
 # Licensed under the GPLv3 or later.
 #
 
-
 # asyncio needs to be imported first to ensure that the reactor is
 # installed properly. Do not re-order.
 import asyncio
 from twisted.internet import asyncioreactor
 asyncioreactor.install(asyncio.get_event_loop())
 
+from pyrdp.core import settings
+from pyrdp.logging import LOGGER_NAMES, NotifyHandler, configure as configureLoggers
 from pyrdp.player import HAS_GUI
-from pyrdp.logging import LOGGER_NAMES, NotifyHandler
+from pyrdp.player.config import DEFAULTS
 
 from pathlib import Path
 import argparse
@@ -28,39 +29,18 @@ if HAS_GUI:
     from PySide2.QtWidgets import QApplication
 
 
-def prepareLoggers(logLevel: int, outDir: Path, headless: bool):
-    logDir = outDir / "logs"
-    logDir.mkdir(exist_ok = True)
+def enableNotifications(logger):
+    """Enable notifications if supported."""
+    # https://docs.python.org/3/library/os.html
+    if os.name != "nt":
+        notifyHandler = NotifyHandler()
+        notifyHandler.setFormatter(logging.Formatter("[{asctime}] - {message}", style = "{"))
 
-    textFormatter = logging.Formatter("[{asctime}] - {levelname} - {name} - {message}", style = "{")
-    notificationFormatter = logging.Formatter("[{asctime}] - {message}", style = "{")
+        uiLogger = logging.getLogger(LOGGER_NAMES.PLAYER_UI)
+        uiLogger.addHandler(notifyHandler)
+    else:
+        logger.warning("Notifications are not supported on this platform.")
 
-    streamHandler = logging.StreamHandler()
-    streamHandler.setFormatter(textFormatter)
-
-    fileHandler = logging.handlers.RotatingFileHandler(logDir / "player.log")
-    fileHandler.setFormatter(textFormatter)
-
-    pyrdpLogger = logging.getLogger(LOGGER_NAMES.PYRDP)
-    pyrdpLogger.addHandler(streamHandler)
-    pyrdpLogger.addHandler(fileHandler)
-    pyrdpLogger.setLevel(logLevel)
-
-    if not headless and HAS_GUI:
-        # https://docs.python.org/3/library/os.html
-        if os.name != "nt":
-            try:
-                notifyHandler = NotifyHandler()
-                notifyHandler.setFormatter(notificationFormatter)
-
-                uiLogger = logging.getLogger(LOGGER_NAMES.PLAYER_UI)
-                uiLogger.addHandler(notifyHandler)
-            except Exception:
-                # No notification daemon or DBus, can't use notifications.
-                pass
-        else:
-            pyrdpLogger.warning("Notifications are not supported for your platform, they will be disabled.")
-    return pyrdpLogger
 
 def main():
     """
@@ -72,15 +52,26 @@ def main():
     parser.add_argument("-b", "--bind", help="Bind address (default: 127.0.0.1)", default="127.0.0.1")
     parser.add_argument("-p", "--port", help="Bind port (default: 3000)", default=3000)
     parser.add_argument("-o", "--output", help="Output folder", default="pyrdp_output")
-    parser.add_argument("-L", "--log-level", help="Log level", default="INFO", choices=["INFO", "DEBUG", "WARNING", "ERROR", "CRITICAL"], nargs="?")
+    parser.add_argument("-L", "--log-level", help="Log level", default=None, choices=["INFO", "DEBUG", "WARNING", "ERROR", "CRITICAL"], nargs="?")
+    parser.add_argument("-F", "--log-filter", help="Only show logs from this logger name (accepts '*' wildcards)", default=None)
     parser.add_argument("--headless", help="Parse a replay without rendering the user interface.", action="store_true")
-
     args = parser.parse_args()
-    outDir = Path(args.output)
-    outDir.mkdir(exist_ok = True)
 
-    logLevel = getattr(logging, args.log_level)
-    logger = prepareLoggers(logLevel, outDir, args.headless)
+    cfg = settings.load(f'{settings.CONFIG_DIR}/player.ini', DEFAULTS)
+
+    # Modify configuration with switches.
+    if args.log_level:
+        cfg.set('vars', 'level', args.log_level)
+    if args.log_filter:
+        cfg.set('logs', 'filter', args.log_filter)
+    if args.output:
+        cfg.set('vars', 'output_dir', args.output)
+
+    configureLoggers(cfg)
+    logger = logging.getLogger(LOGGER_NAMES.PYRDP)
+
+    if cfg.getboolean('logs', 'notifications', fallback=False) and not args.headless:
+        enableNotifications(logger)
 
     if not HAS_GUI and not args.headless:
         logger.error('Headless mode is not specified and PySide2 is not installed. Install PySide2 to use the graphical user interface.')
--- a/pyrdp/core/settings.py
+++ b/pyrdp/core/settings.py
@@ -1,0 +1,33 @@
+#
+# This file is part of the PyRDP project.
+# Copyright (C) 2020 GoSecure Inc.
+# Licensed under the GPLv3 or later.
+#
+from configparser import ConfigParser, ExtendedInterpolation
+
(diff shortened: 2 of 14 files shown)


mitm.default.ini
```diff
@@ -0,0 +1,165 @@
...
+# Formatters
+# -----------------------------------------------------------------
+# Textual / Default Format
+[logs:formatters:default]
+() = pyrdp.logging.formatters.VariableFormatter
+fmt = [{asctime}] - {levelname} - {sessionID} - {name} - {message}
+style = {
+
+[logs:formatters:default:defaultVariables]
+sessionID = GLOBAL
```
Res260: maybe write a comment since users dont know they shouldnt modify this

bin/pyrdp-player.py
```diff
@@ -21,36 +18,26 @@
...
-
-    textFormatter = logging.Formatter("[{asctime}] - {levelname} - {name} - {message}", style = "{")
-    notificationFormatter = logging.Formatter("[{asctime}] - {message}", style = "{")
-
-    streamHandler = logging.StreamHandler()
-    streamHandler.setFormatter(textFormatter)
-
-    fileHandler = logging.handlers.RotatingFileHandler(logDir / "player.log")
-    fileHandler.setFormatter(textFormatter)
+LOG = logging.getLogger(LOGGER_NAMES.PYRDP)
```
obilodeau: We are using global loggers in programs under `bin/` and instance loggers in the core? Is that done consistently?

bin/pyrdp-player.py
```diff
@@ -61,15 +48,24 @@ def main():
     parser.add_argument("replay", help="Replay files to open on launch (optional)", nargs="*")
     parser.add_argument("-b", "--bind", help="Bind address (default: 127.0.0.1)", default="127.0.0.1")
     parser.add_argument("-p", "--port", help="Bind port (default: 3000)", default=3000)
-    parser.add_argument("-o", "--output", help="Output folder", default="pyrdp_output")
-    parser.add_argument("-L", "--log-level", help="Log level", default="INFO", choices=["INFO", "DEBUG", "WARNING", "ERROR", "CRITICAL"], nargs="?")
-
+    parser.add_argument("-o", "--output", help="Output folder", default=None)
+    parser.add_argument("-L", "--log-level", help="Log level", default=None, choices=["INFO", "DEBUG", "WARNING", "ERROR", "CRITICAL"], nargs="?")
+    parser.add_argument("-F", "--log-filter", help="Only show logs from this logger name (accepts '*' wildcards)", default=None)
```
obilodeau: Were these changes reflected in the twistd plugin?

mitm.default.ini
```diff
@@ -0,0 +1,167 @@
...
+# located in the system's user configuration directory and settings
+# will be used unless overridden at the command line.
+#
+# On most Linux distributions, this file should be in
+#
+#     $HOME/.config/pyrdp/mitm.ini
+#
+# On Windows systems, the file should be in one of:
+#
+#     %APPDATA%/<AppName>
```
obilodeau: Aligning examples with the linux version which specifies both path and filename:
```suggestion
# On Windows systems, the file should be:
#
#     %APPDATA%/pyrdp/mitm.ini
```

mitm.default.ini
```diff
@@ -0,0 +1,167 @@
...
+# 
+# On MacOS:
+#
+#     ~/Library/Application Support/<AppName>
```
obilodeau: Aligning examples with the linux version which specifies both path and filename:

```suggestion
#     ~/Library/Application Support/pyrdp/mitm.ini
```

mitm.default.ini
```diff
@@ -0,0 +1,167 @@
...
+stream    = ext://sys.stderr
+formatter = ssl
+
+# -----------------------------------------------------------------
+# Formatters
+# -----------------------------------------------------------------
+# Textual / Default Format
+[logs:formatters:default]
+() = pyrdp.logging.formatters.VariableFormatter
+fmt = [{asctime}] - {levelname} - {sessionID} - {name} - {message}
```
obilodeau: Is it normal that the brackets are unbalanced? Missing a closing square bracket.

mitm.default.ini
```diff
@@ -0,0 +1,167 @@
...
+style = {
+
+# Default session ID name. Do not remove.
```
obilodeau:
```suggestion
# Default session ID naming strategy. Do not remove.
# Right now we name sessions based on a random name and a large but memorizable digit.
# This was favored over the use of a partial hash because it is easier to remember and discuss.
```

player.default.ini
```diff
@@ -0,0 +1,39 @@
+[vars]
```
obilodeau:
```suggestion
# This is a sample PyRDP Player configuration file.  The file should be
# located in the system's user configuration directory and settings
# will be used unless overridden at the command line.
#
# On most Linux distributions, this file should be in
#
#     $HOME/.config/pyrdp/player.ini
#
# On Windows systems, the file should be:
#
#     %APPDATA%/pyrdp/player.ini
#
# On MacOS:
#
#     ~/Library/Application Support/pyrdp/player.ini
[vars]
```

player.default.ini
```diff
@@ -0,0 +1,39 @@
...
+stream    = ext://sys.stderr
+
+[logs:handlers:player]
+class     = logging.handlers.RotatingFileHandler
+filename  = ${vars:output_dir}/${vars:log_dir}/player.log
+formatter = default
+
+[logs:formatters:default]
+class: logging.Formatter
+format = [{asctime}] - {levelname} - {name} - {message}
```
obilodeau: Same concern about square bracket as above.

pyrdp/core/settings.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def load(path: str, fallback: str) -> ConfigParser:
+    """
+    Retrieve the PyRDP settings from a file
+
+    :param path: The path of the file to load.
+    :param fallback: The fallback configuration.
+
+    :returns: A dict of the settings or None if the file was not present.
```
obilodeau: I don't see how this function can ever return None, also type hints state that it returns a `ConfigParser`. Maybe this docstring is outdated with current implementation?

pyrdp/logging/log.py
```diff
@@ -61,3 +45,92 @@ def warning(*args):
...
+        'a': 1,
+        'b': 2,
+        'sublevel': {
+            'c': 3,
+            'd': 4,
+        }
+    }
+    ````
+    """
+    def get(d: dict, key: str, create=False) -> dict:
```
obilodeau: I think using a DefaultDict might simplify that code but maybe I'm wrong since it seems to be optional (`create` arg). Read about it here: https://stackoverflow.com/a/652226/262624

pyrdp/mitm/cli.py
```diff
@@ -117,61 +115,6 @@ def generateCertificate(keyPath: str, certificatePath: str) -> bool:
     result = os.system("openssl req -newkey rsa:2048 -nodes -keyout %s -x509 -days 365 -out %s -subj \"/CN=www.example.com/O=PYRDP/C=US\" 2>%s" % (keyPath, certificatePath, nullDevicePath))
     return result == 0
 
-
-def prepareLoggers(logLevel: int, logFilter: str, sensorID: str, outDir: Path):
```
obilodeau: This is a satisfying removal

pyrdp/mitm/config.py
```diff
@@ -81,3 +81,95 @@ def fileDir(self) -> Path:
         Get the directory for intercepted files.
         """
         return self.outDir / "files"
+
+"""
+The default MITM configuration settings.
+"""
```
obilodeau:
```suggestion
"""
# Changes need to be kept in sync with mitm.default.ini
```

pyrdp/player/config.py
```diff
@@ -0,0 +1,36 @@
+#
+# This file is part of the PyRDP project.
+# Copyright (C) 2020 GoSecure Inc.
+# Licensed under the GPLv3 or later.
+#
+
+DEFAULTS = """
```
obilodeau:
```suggestion
"""
The default Player configuration settings.
"""
# Changes need to be kept in sync with player.default.ini
DEFAULTS = """
```

obilodeau:
There is a conflict to resolve before this can be merged too.

obilodeau:
Hmm, the conflict is fairly involved because of the headless work and the fact that loggers got mass-renamed in the player. I would advise that the conflict resolution should be visible in the PR and not done during the merge. I think either a rebase or a merge from master in the branch will achieve that.

obilodeau:
Merge commit into integration with manual conflict resolution: 76b5b7a774fdbb

obilodeau:
Integration passed CI: https://github.com/GoSecure/pyrdp/runs/534227539?check_suite_focus=true

mitm.default.ini
alxbl: Huh? Where? I see 2 pairs of square brackets (`[logs:...]` and `[asctime]`.

bin/pyrdp-player.py
Res260: All loggers are global when using the `logging` module, I'm not sure what you mean by instance logger

obilodeau: I mean when the logger's lifetime follows one of an object.

mitm.default.ini
obilodeau: I see it now. My bad.

bin/pyrdp-player.py
alxbl: Since there are no objects for the binary scripts, it's a global logger, but there isn't much difference other than aesthetically  since the logging system is all global and `getLogger` only gives you a thread safe handle to the logger. The only place where it matters is for the connection objects which have custom formatters for their session ID. 

I'll do a quick pass in the code base to ensure that the style is consistent.

mitm.default.ini
alxbl: The comments that you suggest are confusing, because this SessionID setting is the one that is used for logs that are not tied to a specific session. It does not influence the per-session naming strategy, as far as I know. I could add some documentation to clarify that this is the "name for the global session" or something.

pyrdp/core/settings.py
alxbl: docstring is outdated from when I was iterating, nice catch.

pyrdp/logging/log.py
alxbl: It would only really remove the `if key in dict` check and add an arbitrary nesting cap, since I need to split and iterate sub dictionaries during the parsing of the ini file anyway to retrieve the right section.

alxbl:
I rebased on master, resolved conflicts and added deduplicated the configuration files. I'm going to fix the twistd CLI issue once and for all, as part of this, too.

I don't want to document the logging configurations yet, as I'd like to expose other configuration settings in the .ini file before making it an official feature.

obilodeau:
I'm sorry to say it but this needs a mention in the README.

The files are very self-documented so no need to copy content over but there should be a pointer to the existence of the ini files and where to grab them. Since our install workflow involves git this will be very simple. For malboxes, installed via pip+git, this is platform-dependent and complicated...

alxbl:
Done. I guess I'll expose the rest of the configuration sooner rather than later.

obilodeau:
Time to rebase my friend! :rocket:

Issue #116, opened 2019-06-11
TLSv1.3 established between client and MITM isn't decryptable with only CLIENT_RANDOM


Pourliver:
The CLIENT_RANDOM isn't enough to decrypt TLSv1.3 traffic. The following zip contains a pcap with TLSv1.3, the CLIENT_RANDOM, and a TLS debug file.

[debug.zip](https://github.com/GoSecure/pyrdp/files/3276991/debug.zip)

Res260:
This *seems* to be caused by the fact that the format of the `CLIENT_RANDOM` has changed and is not fixed-length anymore, and that `CLIENT_HANDSHAKE_TRAFFIC_SECRET` should be used instead.

This should be a simple change in PyRDP, but I’m unsure how it can be done.

reference: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format

Res260:
Also, wireshark's tls dissector failure message:

```
tls13_load_secret Cannot find CLIENT_HANDSHAKE_TRAFFIC_SECRET, decryption impossible
```

obilodeau:
We merged a downgrade workaround in #193. Since this ticket contains a helpful pcap file, I'll turn it into a feature request and keep it.

PR #193, opened 2020-03-18
Various TLS changes

### LINKED ISSUE #192: PyOpenSSL uses TLS 1.0 to connect to the server
See the attached file in #116 and the file I just attached. The file I attached uses OpenSSL 1.1.1 (which is the latest version).
[tls1.0.pcapng.zip](https://github.com/GoSecure/pyrdp/files/4346795/tls1.0.pcapng.zip)

Since TLS 1.0 and 1.1 are deprecated, it might be a problem in the future.

I couldnt confirm 100% that this is a PyOpenSSL problem.

Res260:
See #192. 

- Enable TLS 1.2 support between MITM and RDP Server
- Lock pyopenssl version to avoid breaking changes
- Disable TLS 1.3 for both client MITM and server MITM to avoid "cannot decrypt after the fact" scenarios.

I believe the current code has probably been copied without thinking about this argument, but it limits the TLS versions that PyRDP can use for TLS. 

For context: for the TLS 1.3 disable: #116 

I have tested on my machine, but further investigation is required. If something is not clear hmu.

Fixes #192. Workaround for #116.


Res260:
After further reading: 

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=784153

`SSLv23_METHOD` is the wildcard tls method and will use any available TLS version in OpenSSL. This probably should stay like this for CLIENT --> MITM. Thoughts @obilodeau ? I'll add a comment because this constant name is pure garbage.

Res260:
Alright ready for review now

Res260:
Small change: Lock PyOpenSSL version to avoid breaking changes ( https://www.pyopenssl.org/en/stable/changelog.html )

Res260:
Last change (i, too, gotta sleep): Disable TLS 1.3 (see main comment for context).

alxbl:
I don't think we should close #116 by disabling support for TLS 1.3. There will come a time where it will become necessary. I agree with disabling TLS 1.3 negotiation while we don't properly support it, but I don't feel that's a fix, only a workaround until we can handle 1.3 properly.

Other than not closing #116, this looks good to me.

Res260:
Yeah I wasnt sure what was the best approach, I thought we should open a new issue for TLS 1.3 support which would be an enhancement issue instead of a bug-ish issue. I can remove the fixes statement though no problem

Res260:
@obilodeau do you want to review or I can merge?

obilodeau:
@Res260 reviewing now

pyrdp/core/ssl.py
```diff
@@ -27,9 +27,16 @@ class ClientTLSContext(ssl.ClientContextFactory):
...
-        context = SSL.Context(SSL.TLSv1_METHOD)
+        # Allow the MITM to connect to an RDP Server with ANY TLS version supported by the installed
+        # OpenSSL version. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=784153
+        # It was removed from OpenSSL, but PyOpenSSL has not changed their constant names yet.
+        context = SSL.Context(SSL.SSLv23_METHOD)
         context.set_options(SSL.OP_DONT_INSERT_EMPTY_FRAGMENTS)
         context.set_options(SSL.OP_TLS_BLOCK_PADDING_BUG)
+
+        # We disable TLS 1.3 because the way to decrypt TLS 1.3 traffic differs from
+        # previous TLS versions and is not yet supported by PyRDP.
```
obilodeau: is it really pyrdp that doesn't support 1.3, I thought it was more OpenSSL's master secret extraction that was lagging behind?

setup.py
```diff
@@ -32,7 +32,7 @@
         'names',
         'pyasn1',
         'pycryptodome',
-        'pyopenssl',
+        'pyopenssl==19.1',
```
obilodeau: They seem to be careful about backward-incompatible changes. Maybe we should pin the major version only, not the whole thing so we would get eventual security fixes?

pyrdp/core/ssl.py
Res260: OpenSSL (starting from v1.1.1) supports TLS 1.3 and the necessary methods to extract the needed secrets/keys (but its different from <TLS1.3). PyOpenSSL doesnt seem to expose them, but I could be wrong. For this reason,  I prefer to say that PyRDP doesnt support TLS 1.3, because I'm not sure if PyOpenSSL is really lagging behind.

Supporting TLS 1.3 would also probably increase the required OpenSSL version to 1.1.1, and i'm not sure of the different package manager's versions of OSSL. More investigation is required

setup.py
Res260: I locked to the minor version `.1`, so I thought it was enough to receive the security updates, but it made me realise that they may not the older minor versions. I'll lock to major.

Res260:
> Also, from now on, we should collect pcaps for TLS issues and add them as integration tests.

This might require more work than it seems, as the current integration test purposely avoids TLS testing.

obilodeau:
> This might require more work than it seems, as the current integration test purposely avoids TLS testing (because neither me nor @xshill used TLS with scapy).

Fair point. I missed that. Well, I'll try if we ever have a regression in the future and give up eventually then :wink:. Let's *not* track that in an issue so we don't feel bad about ourselves :smile:

pyrdp/core/ssl.py
obilodeau: Fair enough.

Res260:
Just thought about it, should this go in the changelogs?

obilodeau:
> Just thought about it, should this go in the changelogs?

I'm doing changelog updates after I merge in a separate push straight into master. This reduces the amount of conflict handling that we have to do.

It's not a problem if you provide it but it's not a problem if you don't. I've been using that process with a lot of success over at [asciidoctor-reveal.js](https://github.com/asciidoctor/asciidoctor-reveal.js).

PR #194, opened 2020-03-19
Add a CI and a basic integration test

Res260:
I had a burst of motivation and did this, I don't really care if you want it or not but offer it nonetheless, I mostly did this as a personnal challenge :)

**Basically, I added a Github Actions workflow that does two things on push and on Pull Request on the repo:**
**1. Try to build the Docker Image.**
**2. Try to install PyRDP and run a new integration test I made on ubuntu and windows.**
**3. Show a coverage report and fail it it falls under 50%**

This test is based on an RDP session I just recorded which try to use as most RDP features that PyRDP supports as possible. 

Right now, it does assertions on these:
- Asserts that PyRDP correctly build the requested virtual channels;
- Asserts that PyRDP sees a clipboard PDU with the pasted data;
- Asserts that the number of mouse clicks parsed matches the number of total clicks;
- Asserts that PyRDP clearly sees some text that was typed on the keyboard at the end of the session;

Beyond those assertions, it executes a large portion of the PyRDP codebase:
image
image

It does **NOT** test these:
- TLS, RC4 and anything else crypto
- Logging
- Replays
- Device Redirection Channel
- Executables in the "bin" folders
- The first two packets of the RDP connection

This test could be better (and can be improved in the future), but I believe it will catch most low-hanging fruits/bugs! 

Because of (what I believe is) a synchronization problem, I disabled testing for the device redirection channel.

It is heavily based on the pyrdp-replay branch (requires scapy as well).


test/test.py
```diff
@@ -0,0 +1,249 @@
...
+                test_mitm.recvFromClient(data)
+            elif source == server_ip and destination == mitm_ip:
+                test_mitm.recvFromServer(data)
+            elif source == mitm_ip and destination == client_ip and destination_port != 3389:
+                test_mitm.sendToClient(data)
+            elif source == mitm_ip and destination == server_ip:
+                test_mitm.sendToServer(data)
+            else:
+                assert False
+        except NotImplementedError as e:
```
alxbl: Isn't this entire try-block redundant? Just let the exception be thrown, unless I'm missing something?

test/test.py
```diff
@@ -0,0 +1,249 @@
...
+            raise e
+
+    try:
+        test_mitm.tcp.recordConnectionClose()
+    except struct.error as e:
```
alxbl: Should this cause the test to fail?
Should you catch a generic exception? What if the code changes and throws a different exception? The test will no longer report failure/unexpected behavior

test/test.py
```diff
@@ -0,0 +1,249 @@
...
+        print("Couldn't close the connection cleanly. "
+              "Are you sure you got source and destination correct?")
+
+    assert test_mitm.builtIOChannel, "PyRDP did not build IO Channel."
+    assert test_mitm.builtCliprdrChannel, "PyRDP did not build the Clipboard Channel."
+    assert test_mitm.builtRDPDRChannel, "PyRDP did not build the RDPDR Channel."
+
+    logging.critical("Channel building assertions PASSED")
```
alxbl: Don't need to log success in integration tests.
Also don't increase severity to ensure something gets printed. This should be INFO, if logged at all. In tests, no output == no problem

test/test.py
```diff
@@ -0,0 +1,249 @@
...
+
+    for key, value in test_mitm.clipboardObserver.expectedClipboardData.items():
+        assert value, f"Expected to receive {key} in a clipboardPDU but the clipboard observer did not receive it."
+
+    logging.critical("Clipboard content assertions PASSED")
```
alxbl: See above

test/test.py
```diff
@@ -0,0 +1,249 @@
...
+    actualLeftMouseClicks = test_mitm.inputObserver.leftMouseClicks
+    expectedRightMouseClicks = 3
+    actualRightMouseClicks = test_mitm.inputObserver.rightMouseClicks
+    expectedMouseWheelClicks = 0
+    actualMouseWheelClicks = test_mitm.inputObserver.mouseWheelClicks
+    assert actualLeftMouseClicks == expectedLeftMouseClicks, f"Wrong number of left mouse clicks registered. Expected {expectedLeftMouseClicks}, got {actualLeftMouseClicks}"
+    assert actualRightMouseClicks == expectedRightMouseClicks, f"Wrong number of right mouse clicks registered. Expected {expectedRightMouseClicks}, got {actualRightMouseClicks}"
+    assert actualMouseWheelClicks == expectedMouseWheelClicks, f"Wrong number of mouse wheel clicks registered. Expected {expectedMouseWheelClicks}, got {actualMouseWheelClicks}"
+
+    logging.critical("Mouse clicks assertions PASSED")
```
alxbl: See above

test/test.py
```diff
@@ -0,0 +1,249 @@
...
+
+    assert "arrrray" in test_mitm.state.inputBuffer, f"'arrrray' not found in the MITM state input buffer, but was typed during the session. Input buffer: '{test_mitm.state.inputBuffer}'"
+
+    logging.critical("Keyboard typing assertion PASSED")
```
alxbl: See above

alxbl:
This is very cool! I think it's a great first step towards having proper unit testing and integration testing in PyRDP. I have a few comments but the general approach looks good to me.

Good use of dependency injection for the configs, too.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Pip version
+      run: pip --version
+
+    - name: Install setuptools
+      run: sudo apt install python3-setuptools
+    - name: Install PyRDP dependencies
+      run: sudo apt install libdbus-1-dev libdbus-glib-1-dev libgl1-mesa-glx git python3-dev
+    - name: Install PyRDP
+      working-directory: .
+      run: pip install -U -e .
```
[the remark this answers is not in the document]
Res260: It should be fine!

test/test.py
Res260: Yeah, probably, it's from the pyrdp-replay script and I didnt remove it but probably should

test/test.py
Res260: Makes sense, will change

test/test.py
Res260: removed it

test/test.py
```diff
@@ -0,0 +1,249 @@
...
+    def __init__(self):
+        super().__init__()
+        self.leftMouseClicks = 0
+        self.rightMouseClicks = 0
+        self.mouseWheelClicks = 0
+
+    def onPDUReceived(self, pdu: FastPathPDU):
+        for event in pdu.events:
+            if isinstance(event, FastPathMouseEvent):
+                if event.pointerFlags & PointerFlag.PTRFLAGS_DOWN and event.pointerFlags & PointerFlag.PTRFLAGS_BUTTON1:
```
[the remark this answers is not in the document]
Res260: done

alxbl:
If you can rebase this and `--force-with-lease` to clean up the commit history, I'd be willing to merge this now.

Res260:
@alxbl squash and merge would do the job?

obilodeau:
@alxbl I would like to review before you merge, please. I'm very interested in GitHub Actions beyond just this repo.

alxbl:
> @alxbl wouldn't squash and merge do the job?
> 
> Edit: wording as the first message seemed passive-agressive

It will, but your commits will disappear. I was offering you to keep your commit history in master.

Res260:
Ahhh I don't really care about having multiple commits for this PR, and the branch is still kept in GitHub if we really need the information

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build Docker image
+      run: docker build .
+
+  Install-and-test:
```
obilodeau:
```suggestion
  install-and-test:
```
for consistency with other names

pyrdp/mitm/BasePathMITM.py
```diff
@@ -60,4 +60,4 @@ def onScanCode(self, scanCode: int, isReleased: bool, isExtended: bool):
         # Normal input
         elif len(keyName) == 1:
             if not isReleased:
-                self.state.inputBuffer += keyName
\ No newline at end of file
+                self.state.inputBuffer += keyName
```
obilodeau: Please refrain from unnecessary line-noise. Learn to use `git add -p` or your IDE's equivalent.

pyrdp/mitm/RDPMITM.py
```diff
@@ -37,15 +37,16 @@
...
+    Recorder
 
 
 class RDPMITM:
     """
     Main MITM class. The job of this class is to orchestrate the components for all the protocols.
     """
 
-    def __init__(self, mainLogger: SessionLogger, crawlerLogger: SessionLogger, config: MITMConfig):
+    def __init__(self, mainLogger: SessionLogger, crawlerLogger: SessionLogger, config: MITMConfig, state: RDPMITMState=None, recorder: Recorder=None):
```
obilodeau: Note to self: this is duplicated from another active PR

test/test.py
```diff
@@ -0,0 +1,243 @@
...
+from pyrdp.mitm.MITMRecorder import MITMRecorder
+from pyrdp.mitm.state import RDPMITMState
+from pyrdp.pdu import ClipboardPDU, FastPathMouseEvent, FastPathPDU, FormatDataResponsePDU
+
+
+def bytesToIP(data: bytes):
+    return ".".join(str(b) for b in data)
+
+
+def parseExportedPdu(packet: packet.Raw):
```
obilodeau: Code duplicated from an active PR. One of us will rebase and refactor to avoid this useless duplication.

obilodeau:
A simple zip of the pcap gave a 5.9M binary (from 16M). It's not as good as I have expected. Since git stores it's pack metadata already compressed (so also on the network), I'm going to let go on this.

I might revisit if we add more pcap-based tests to reduce checkout size.

.github/workflows/ci.yml
```diff
@@ -36,4 +36,4 @@ jobs:
 
     - name: Integration Test with a prerecorded PCAP.
       working-directory: ./
-      run: python test/test.py
\ No newline at end of file
+      run: python test/test.py
```
obilodeau: again! ;)

Res260:
> A simple zip of the pcap gave a 5.9M binary (from 16M). It's not as good as I have expected. Since git stores it's pack metadata already compressed (so also on the network), I'm going to let go on this.
> 
> I might revisit if we add more pcap-based tests to reduce checkout size.

Should I revert the commit then?

pyrdp/mitm/BasePathMITM.py
Res260: but pep8 :(

obilodeau:
> Should I revert the commit then?

Since it's already done I'm tempted to say no.

That said, are you sure this directory will always be writable to the user running the tests? It looks like the GitHub action worked so we can keep it.

Res260:
> That said, are you sure this directory will always be writable to the user running the tests? It looks like the GitHub action worked so we can keep it.

There is no garantee indeed, i'm not sure what would be the best approach. It will always work for the CI though

Res260:
For the code coverage, I could add it to the CI if you want using https://www.codecov.io/, thoughts?

.github/workflows/ci.yml
Res260: Not sure I understand 😂

Res260:
Actually I can do it inside the CI without a third party

pyrdp/mitm/BasePathMITM.py
obilodeau: Pep8 says files should end with a `\n` (like vim does), I'm all for it.

Reducing line noise is about making it easier to review for the reviewer by removing all the noise that has nothing to do with the PR. If you touch a function or create a new file, please follow PEP8 but if you do PEP8 enforcement changes in the same commit / PR as another feature then it makes review harder to do.

When an existing code-base already has lacking PEP8 enforcement, using a tool to convert it all to PEP8 compliant is another case of noise that is added for little benefit. Most of the time when trying to find commit introducing problems with `git log -p` reverse search, you will always stumble on these big "re-indent" commits and it slows down your quest. For such codebases the re-indent / pep8 enforcement should be done gradually as you are modifying the code for another purpose.

Of course, this is a mix of experience and personal preference, your mileage may vary.

obilodeau:
> For the code coverage, I could add it to the CI if you want using https://www.codecov.io/, thoughts?

then

> Actually I can do it inside the CI without a third party

Do what you prefer.

Res260:
Added windows installation and test, as a treat :)

It does the same thing as the ubuntu one but on windows. My reasonning behind this is that even though installation does not fail per se, execution can fail because dependencies sometime have a hard time testing for both windows and linux, so I'd rather run the test twice.

Res260:
I added the coverage report to the CI, and it will fail if it falls under 50% (currently 53% coverage).

@obilodeau Should I still add documentation for the coverage in docs or it's not needed?

obilodeau:
> @obilodeau Should I still add documentation for the coverage in docs or it's not needed?

Not needed. This is self-explanatory now. Thanks!

PR #195, opened 2020-03-19
feat: Added --no-files switch to disable  file carving.

alxbl:
Adds a `--no-files` switch that disables interception of file transfers between the client and server.

### DIFF
--- a/pyrdp/mitm/DeviceRedirectionMITM.py
+++ b/pyrdp/mitm/DeviceRedirectionMITM.py
@@ -77,7 +77,7 @@ class DeviceRedirectionMITM(Subject):
         self.fileMapPath = self.config.outDir / "mapping.json"
         self.forgedRequests: Dict[int, DeviceRedirectionMITM.ForgedRequest] = {}
 
-        if state.config.extractFiles:
+        if self.config.extractFiles:
             self.responseHandlers: Dict[MajorFunction, callable] = {
                 MajorFunction.IRP_MJ_CREATE: self.handleCreateResponse,
                 MajorFunction.IRP_MJ_READ: self.handleReadResponse,
@@ -328,7 +328,7 @@ class DeviceRedirectionMITM(Subject):
         """
 
         if not self.config.extractFiles:
-            self.log.warning('Ignored attempt to forge file reads because file extraction is disabled.')
+            self.log.info('Ignored attempt to forge file reads because file extraction is disabled.')
             return
 
         self.statCounter.increment(STAT.DEVICE_REDIRECTION_FORGED_FILE_READ)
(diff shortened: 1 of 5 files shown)


pyrdp/mitm/DeviceRedirectionMITM.py
```diff
@@ -70,19 +70,21 @@ def __init__(self, client: DeviceRedirectionLayer, server: DeviceRedirectionLaye
...
         self.fileMap: Dict[str, FileMapping] = {}
         self.fileMapPath = self.config.outDir / "mapping.json"
         self.forgedRequests: Dict[int, DeviceRedirectionMITM.ForgedRequest] = {}
 
-        self.responseHandlers: Dict[MajorFunction, callable] = {
-            MajorFunction.IRP_MJ_CREATE: self.handleCreateResponse,
-            MajorFunction.IRP_MJ_READ: self.handleReadResponse,
-            MajorFunction.IRP_MJ_CLOSE: self.handleCloseResponse,
-        }
+        if state.config.extractFiles:
```
obilodeau: The `config()` `@property` was not used here to access the config. Any reason why? Maybe because it is `__init__` and it wouldn't work?

pyrdp/mitm/DeviceRedirectionMITM.py
```diff
@@ -325,6 +327,10 @@ def sendForgedFileRead(self, deviceID: int, path: str) -> int:
         :param path: path of the file to download. The path should use '\' instead of '/' to separate directories.
         """
 
+        if not self.config.extractFiles:
+            self.log.warning('Ignored attempt to forge file reads because file extraction is disabled.')
```
obilodeau: I think warning is a too intense log level for this. Since the feature was activated by the user, info would suffice.

pyrdp/mitm/DeviceRedirectionMITM.py
```diff
@@ -345,6 +350,10 @@ def sendForgedDirectoryListing(self, deviceID: int, path: str) -> int:
         \Documents\*
         """
 
+        if not self.config.extractFiles:
+            self.log.warning('Ignored attempt to forge directory listing because file extraction is disabled.')
```
obilodeau: Same as above comment: warning too intense, info would be better

pyrdp/mitm/DeviceRedirectionMITM.py
obilodeau:
```suggestion
            self.log.info('Ignored attempt to forge file reads because file extraction is disabled.')
```

pyrdp/mitm/DeviceRedirectionMITM.py
obilodeau:
```suggestion
            self.log.info('Ignored attempt to forge directory listing because file extraction is disabled.')
```

pyrdp/mitm/DeviceRedirectionMITM.py
alxbl: Mostly because  the state object is available directly, but I'll change it to be consistent.

PR #197, opened 2020-03-23
Add support for transparent proxying on Linux

alxbl:
This feature adds the ability to tell PyRDP to perform a transparent connection to the server by using the source IP of the client. This cannot be fully supported by PyRDP and relies on firewall configuration to properly route traffic.

The PR adds documentation of an example scenario and configuration, along with the required PyRDP code to create a transparent socket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ PyRDP was [introduced in 2018](https://www.gosecure.net/blog/2018/12/19/rdp-man-
       - [Choosing when to resume normal activity](#choosing-when-to-resume-normal-activity)
     + [Other MITM arguments](#other-mitm-arguments)
       - [--no-downgrade](#--no-downgrade)
+      - [--transparent](#--transparent)
   * [Using the PyRDP Player](#using-the-pyrdp-player)
     + [Playing a replay file](#playing-a-replay-file)
     + [Listening for live connections](#listening-for-live-connections)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -331,6 +332,60 @@ to be established. The following are currently not affected by this switch and w
...
+SERVER_IP=10.2.2.2
+
+# The mark number to use in iptables (should be fine as-is)
+MARK=1
+
+# The routing table ID for custom rules (should be fine as-is)
+TABLE_ID=100
+
+# Create a custom routing table for pyrdp traffic
+echo "$TABLE_ID    pyrdp" >> /etc/iproute2/rt_tables
```
Res260: maybe format using ```console

pyrdp/core/twisted.py
```diff
@@ -29,4 +34,39 @@ def buildProtocol(self, addr):
...
+
+class TransparentConnector(tcp.Connector):
+    """A TCP connector which creates TransparentClients."""
+
+    def _makeTransport(self):
+        return TransparentClient(self.host, self.port, self.bindAddress, self, self.reactor)
+
+
+def connectTransparent(host, port, factory, timeout=30, bindAddress=None):
+    from twisted.internet import reactor
```
Res260: why is the import statement there?

Res260:
Very cool 👍

pyrdp/core/twisted.py
alxbl: I wanted to make sure that importing this file first does not initialize a different reactor than the one installed in `pyrdp-mitm.py`. Since twisted seems to lazily create the reactor. I might be wrong though, but I remember that moving some imports related to twisted earlier caused some exceptions. So I'd rather lazily import reactor the first time this function is called.

alxbl:
> should be documented as incompatible with docker images. 

I remember writing something about that. I can't remember if it was in the code or decided to not include it after, but yes, I agree. I'll clarify.

> what about after? When the process replies to the original host, shouldn't a transparent socket be used there also? Otherwise you can detect in a packet capture that your server is not the one you expect. Am I missing something?

The packets coming from the target server (replies) are marked using a PREROUTING rule. Then an ip routing rule is saying that all of the packets marked with that mark must be routed using a custom routing table, which contains a single rule: `route through loopback`. This effectively forces the packets to hit loopback, and thus the transparent socket opened during the connection receives the data. PyRDP has all of the state to forward back to the client socket.

It is kind of mind-bending, but it works :)

obilodeau:
I think I wasn't clear enough. I don't see how packets coming back from the MITM to the client are being source rewritten so that to the client it's the real server that replied and not the MITM. I'll diagram it:

```
             DNAT happens here  +  transparent socket is here
                  +             |
                  |             |
+--------+        v  +--------+ v        +---------+
| Client +--------+->+  MITM  +-+-------->  Server |
|        |           |        |          |         |
|        +<-+--------+        +<---------+         |
+--------+  ^        +--------+          +---------+
            |
            |
            + I don't see how this flow could have the right
              source IP (10.2.2.2). Nothing in the iptable
              rules or the code does anything to do so.
```

If you don't have a pcap from the client's perspective, I'll checkout the branch and try to create one to verify. Or maybe you can answer me directly too.

alxbl:
I'm pretty sure it's because the socket is already opened with the intended destination of 10.2.2.2 when being prerouted, so the source IP on replies is set to 10.2.2.2 by the MITM's linux kernel.

here's a PCAP: [proxy.tar.gz](https://github.com/GoSecure/pyrdp/files/4388378/proxy.tar.gz)
10.6.66.24: Client (priv0)
10.2.0.126: Server (qemu0)
10.2.0.1: MITM (Gateway) (qemu0)

PR #198, opened 2020-03-25
Dockerfile improvements again

obilodeau:
Since we merged #190 we can apply further sizing improvements to our docker image

* [x] Split GUI and not GUI into two images
* [x] Staged build to drastically reduce image size (no more `build-essentials` package required)
* [x] verify docker size improvements gains
* [x] Apply staged build to GUI Dockerfile
* [x] Add ~`integration-gui`~ `integration-slim` and ~`latest-gui`~ `latest-slim` tags to dockerhub automatic builds
* [x] update `docker-compose.yml` and documentation
* [x] add a `Dockerfile.slim` build in CI too

### Size reduction
* The integration image is 63.24 MB now. That is down from 252.39 MB now in master or down from 570.26 MB for 0.4.1.
* The integration-gui image is 293.32 MB now. That is up from 252.39 MB now in master (which I suspect was broken for the GUI since I needed to add a package to make it work on my system) or down from 570.26 MB for 0.4.1.

### Feedback required for Tag names
Because of @alxbl's upcoming work with video encoding and the needed dependencies, I'm not sure that `-gui` is the proper approach. Candidates:
* a) Current proposal: `latest` with minimal dependencies and `latest-gui` (with GUI but more later?)
* b) `latest` with minimal dependencies and `latest-full` (with everything)
* c) Turn it the other way around: `latest` with everything (GUI + video encoding, etc.) and `latest-slim` (I think python does it that way for official images)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,16 @@
 # Handles compiling and package installation
 FROM ubuntu:18.04 AS compile-image
-# Install Dependencies
+# Install build dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 python3-pip \
         # Required for local pip install
         python3-setuptools \
         # Required for venv setup
         python3-venv \
-        # Required to build RLE module
-        build-essential python3-dev
+        # Required to build RLE module and dbus-python
+        build-essential python3-dev \
+        libdbus-1-dev \
+        libdbus-glib-1-dev
 
 RUN python3 -m venv /opt/venv
 # Make sure we use the virtualenv:
@@ -21,14 +23,18 @@ COPY ext/rle.c /pyrdp/ext/rle.c
 COPY pyrdp/ /pyrdp/pyrdp/
 # Install in the virtualenv
 RUN cd /pyrdp \
-    && pip3 --no-cache-dir install . -U
+    && pip3 --no-cache-dir install .[GUI] -U
+
 
 # Handles runtime only (minimize size for distribution)
 FROM ubuntu:18.04 AS docker-image
 
-# Install Dependencies
+# Install runtime dependencies except pre-built venv
 RUN apt-get update && apt-get install -y --no-install-recommends python3 \
-    && rm -rf /var/lib/apt/lists/*
+        # GUI and notifications stuff
+        libgl1-mesa-glx \
+        notify-osd dbus-x11 libxkbcommon-x11-0 \
+        && rm -rf /var/lib/apt/lists/*
 
 # Copy preinstalled dependencies from compile image
 COPY --from=compile-image /opt/venv /opt/venv
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,44 @@
-FROM ubuntu:18.04
-
+# Handles compiling and package installation
+FROM ubuntu:18.04 AS compile-image
 # Install Dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 python3-pip \
         # Required for local pip install
         python3-setuptools \
-        # Required to build dbus-python
-        build-essential python3-dev \
-        libdbus-1-dev \
-        libdbus-glib-1-dev \
-        libgl1-mesa-glx \
-        notify-osd dbus-x11 \
+        # Required for venv setup
+        python3-venv \
+        # Required to build RLE module
+        build-essential python3-dev
+
+RUN python3 -m venv /opt/venv
+# Make sure we use the virtualenv:
+ENV PATH="/opt/venv/bin:$PATH"
+
+# Copy only what is required for the install
+COPY setup.py /pyrdp/setup.py
+COPY bin/ /pyrdp/bin/
+COPY ext/rle.c /pyrdp/ext/rle.c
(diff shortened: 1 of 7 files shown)


obilodeau:
New Dockerfile worked locally. Triggering an integration build to see how it goes over there and it's resulting size.

Dockerfile
```diff
@@ -1,25 +1,44 @@
...
         python3-setuptools \
-        # Required to build dbus-python
-        build-essential python3-dev \
-        libdbus-1-dev \
-        libdbus-glib-1-dev \
-        libgl1-mesa-glx \
-        notify-osd dbus-x11 \
-    && rm -rf /var/lib/apt/lists/*
+	# Required for venv setup
+	python3-venv \
```
obilodeau: whitespace issues

obilodeau:
Verified image size in integration. Added at the top of the issue.

> The integration image is 63.24 MB now. That is down from 252.39 MB now in master or down from 570.26 MB for 0.4.1.

Also added a question about tag names alternatives

obilodeau:
Verified image size in integration-gui. Added at the top of the issue.

> The integration-gui image is 293.32 MB now. That is up from 252.39 MB now in master (which I suspect was broken for the GUI since I needed to add a package to make it work on my system) or down from 570.26 MB for 0.4.1.

obilodeau:
@alxbl when you have a minute, let's discuss the tag name approach that I added at the top of the PR. @Res260 feel free to chime in as well.

Once this will be figured out, I'll setup dockerhub and update the documentation accordingly.

alxbl:
> @alxbl when you have a minute, let's discuss the tag name approach that I added at the top of the PR. 

I feel like a lot of images take the `-slim` tag approach, so I kinda like that one. By default you get everything (batteries included), and if you have specific needs, then you read the docs and find out that there's a slim build or a specific tag for your needs.

obilodeau:
Changed the Dockerfile to use the slim approach. Configured an integration-slim build rule in DockerHub.

obilodeau:
Image builds that happened during lunch:
* integration: 295.44 MB
* integration-slim: 65.37 MB

obilodeau:
Merged this into integration to see if it builds. I'll finish and submit the PR tomorrow.

Res260:
@obilodeau I don't know if you wanted to do it like that, but you can separate the two docker builds in two different jobs in the CI so they can run in parrallel instead of sequentially

obilodeau:
> @obilodeau I don't know if you wanted to do it like that, but you can separate the two docker builds in two different jobs in the CI so they can run in parrallel instead of sequentially

I decided to keep it into the same task consciously. My rationale was that this is one less ubuntu to start so we save a few seconds of CI time (not sure how much we use though). The two builds are closely related and some of the docker layers might be locally cached. I also believe that if one fails the other will probably too.

I haven't thought of the parallel argument. Good call. If I ever find myself waiting too much on this specific CI task I'll think about it.

README.md
```diff
@@ -200,6 +207,12 @@ First of all, build the image by executing this command at the root of PyRDP (wh
 docker build -t pyrdp .
 ```
 
+As an alternative we have a slimmer image without the GUI and ffmpeg dependencies:
+
+```
+docker build -f Dockerfile.slim -t pyrdp .
```
Res260, line 213: is it normal to have this same text at two places in the readme?

docker-compose.yml
```diff
@@ -3,7 +3,17 @@
...
 services:
   pyrdp:
-    build: .
+    image: gosecure/pyrdp:latest
+    # Slim image (no GUI pyrdp-player, no pyrdp-convert)
+    #image: gosecure/pyrdp:latest-slim
+    # Uncomment a build section if you want to build your own image
+    # Full image
+    #build: .
+    # Slim image (no pyrdp-player, no pyrdp-convert)
```
Res260, line 12: is this still true? I thought the slim image could run pyrdp-player in headless mode

README.md
obilodeau: One is in the section to install pyrdp (`docker pull`) and the other is in the section if you want to build pyrdp with docker (`docker build`). Installing is described on top of the README and advanced usage like docker builds is described later in the file. I think it's normal and fine.

docker-compose.yml
obilodeau:
```suggestion
    # Slim image (no pyrdp-player w/o --headless and no pyrdp-convert)
```

better?

Res260: yeah, though I might be nitpicking too much but I find w/o to be confusing and unintuitive, I think shortcuts like that should be avoided in documentations

obilodeau:
```suggestion
    # Slim image (no pyrdp-player without --headless and no pyrdp-convert)
```
now?

README.md
obilodeau: Can I resolve this?

docker-compose.yml
Res260: yes

README.md
Res260: yeah, doing it

PR #201, opened 2020-03-25
Remove CI trigger on push, keep only on PRs and master pushes

Res260:
This avoids the "double, identical CI runs" problem, and since we only work with PRs, it's not a problem to remove it. I kept CI run triggers on pushes to `master`


Res260:
updated the PR, edited title and main comment

PR #205, opened 2020-04-06
fix: Add TPROXY support for client-MITM connections

### LINKED ISSUE #64: Add images to the readme
- Player
- Console output from the MITM

Helps with quick glances at the project.

### LINKED ISSUE #204: Layer 2 transparent proxying
We toyed with the idea of doing L2 transparent proxying instead of L3. This would be more flexible if you need to drop a device on a network segment with DHCP.

It took us more time than I'm willing to admit but here are the rough steps.

```
  +--------+           +------+             +--------+
  | CLIENT | <-- 1 --> | MITM | <--- 2 ---> | SERVER |
  +--------+           +------+             +--------+
 10.13.37.10         10.13.37.118*          10.13.37.111
  
*: Assigning an IP to the MITM is necessary for operation
```

Netplan bridge configuration:
```yaml
network:
  version: 2
  renderer: networkd
  ethernets:
    enp0s3:
      dhcp4: no
    enp0s8:
      dhcp4: no
  bridges:
    br0:
      dhcp4: yes
      # or
      #addresses:
        #- 10.13.37.11/24
      interfaces:
        - enp0s3
        - enp0s8
```

In my case, `enp0s3` is the interface on the MITM system with is on the segment shown as (1) above. `enp0s8` is the interface on the MITM system on the segment shown as (2) above.

`setup.sh:`:
```shell
#!/bin/bash -x

# The IP of the RDP server which will receive proxied connections.
SERVER_IP=10.13.37.111

# The mark number to use in iptables (should be fine as-is)
MARK=1

# The routing table ID for custom rules (should be fine as-is)
TABLE_ID=100

# Create a custom routing table for pyrdp traffic
echo "$TABLE_ID    pyrdp" >> /etc/iproute2/rt_tables

# Route RDP traffic intended for the target server to local PyRDP (1)
iptables -t nat \
    -A PREROUTING \
    -d $SERVER_IP \
    -p tcp -m tcp --dport 3389 \
    -j REDIRECT --to-port 3389

# Mark RDP traffic intended for clients (2)
iptables -t mangle -A PREROUTING \
    -s $SERVER_IP \
    -m tcp -p tcp --sport 3389 \
    -j MARK --set-mark $MARK

# Set route lookup to the pyrdp table for marked packets.
ip rule add fwmark $MARK lookup $TABLE_ID

# Add a custom route that redirects traffic intended for the outside world to loopback
# So that server-client traffic passes through PyRDP
# This table will only ever be used by RDP so it should not be problematic
ip route add local default dev lo table $TABLE_ID

# If you want the interception to happen at the Layer 2
# Make sure that both your interfaces are bridged together

# WARNING: if you have other important firewall rules, make sure to test the impact of this
modprobe br_netfilter
echo 1 > /proc/sys/net/bridge/bridge-nf-call-iptables

# Target DROP in brouting means out of L2 go route in L3
ebtables -t broute -A BROUTING -i enp0s8 -p ipv4 --ip-src $SERVER_IP --ip-proto tcp --ip-source-port 3389 -j redirect --redirect-target DROP
# recent linux like debian buster needs to use ebtables-legacy as nft ebtables doesn't support the broute table yet
#ebtables-legacy -t broute -A BROUTING -i enp0s8 -p ipv4 --ip-src $SERVER_IP --ip-proto tcp --ip-source-port 3389 -j redirect --redirect-target DROP

# Disable return path filtering
echo 0 > /proc/sys/net/ipv4/conf/default/rp_filter
echo 0 > /proc/sys/net/ipv4/conf/all/rp_filter
echo 0 > /proc/sys/net/ipv4/conf/enp0s3/rp_filter
echo 0 > /proc/sys/net/ipv4/conf/enp0s8/rp_filter
```

`cleanup.sh` (Warning: it is a bit aggressive on firewall rule deletion, adapt to your needs):
```shell
#!/bin/bash -x
# Use same value as in setup script
MARK=1
TABLE_ID=100

# Remove custom routing table for pyrdp traffic
sed -i '/pyrdp/d' /etc/iproute2/rt_tables

# Remove RDP traffic redirection
iptables -t nat \
    -F PREROUTING

# Remove all marking
iptables -t mangle -F PREROUTING

# Remove route lookup for marked packets
ip rule del fwmark $MARK lookup $TABLE_ID

# Remove custom route that redirects outside traffic to loopback
ip route del local default dev lo table $TABLE_ID

# Remove L2 to L3 packet passing
ebtables -t broute -F BROUTING

# Reseting kernel defaults
# bridging shouldn't hit iptables rules
echo 0 > /proc/sys/net/bridge/bridge-nf-call-iptables
# re-activating return path filtering
echo 1 > /proc/sys/net/ipv4/conf/default/rp_filter
echo 1 > /proc/sys/net/ipv4/conf/all/rp_filter
echo 1 > /proc/sys/net/ipv4/conf/enp0s3/rp_filter
echo 1 > /proc/sys/net/ipv4/conf/enp0s8/rp_filter
```

Invoking pyrdp (as root!):
```shell
pyrdp-mitm.py --transparent 10.13.37.111
```

I'll merge this with the upstream docs eventually but not now. I wanted to keep track of it so here it is.

*: This is a limitation that we should be able to remove but I couldn't figure it out in a timely fashion.

alxbl:
This PR implements a clean version of the patch documented in #204. It also splits the transparent proxying configuration examples to a different file to keep the main README clean.

I'd like @obilodeau to merge his documentation from #204 in `docs/transparent-proxy.md` and then we can close that issue as well.

---
Closes #204, Closes #64


README.md
```diff
@@ -336,55 +336,12 @@ it would then become possible to extract a valid RDP replay file from the raw ne
...
-
-This setup is a base example and might be much more complex depending on your environment and constraints.
-To cleanup, you should delete the created routes, remove the custom routing table from `rt_tables` and
-remove the `ip rule`.
+address instead of the MITM one. This option is only useful in certain scenarios where the MITM is physically
+a gateway between clients and the server and sees all traffic. 
+[Specific examples can be found here.](docs/transparent-proxy.md)
+
+**NOTE**: This requires root privileges, only works on Linux and requires manual firewall configuration to ensure 
+that traffic is router properly.
```
Res260, line 344: routed?

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+  bridges:
+    br0:
+      interfaces:
+        - enp0s3
+        - enp0s8
+      # if you are confident that your deployment is not going
+      # to cause a loop you should leave these enabled, otherwise comment them
+      parameters:
+        stp: false
+        forward-delay: 0
```
obilodeau, line 89: This can be removed which makes the whole idea of documenting with netplan pointless. I would start the docs with a single management interface using DHCP and setup the bridge manually using network namespace right from the start.

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+
+**setup.sh**
+
+```shell
+#!/bin/bash -x
+#               enp0s3            enp0s8
+#  +--------+         \ +------+ /           +--------+
+#  | CLIENT | <-- 1 --> | MITM | <--- 2 ---> | SERVER |
+#  +--------+           +------+             +--------+
+#  10.13.37.10             br0              10.13.37.111
```
obilodeau, line 100: nice additions to the asciiart :slightly_smiling_face:

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+# WARNING: unclear if required on not
+echo 1 > /proc/sys/net/ipv4/conf/br0/route_localnet
+echo 1 > /proc/sys/net/ipv4/conf/enp0s3/route_localnet
+
+# Mark socket Packets
+# WARNING: Unclear if required or not but likely good for performance so it will likely stay
+iptables -t mangle -N DIVERT
+iptables -t mangle -A DIVERT -j MARK --set-mark 1
+iptables -t mangle -A DIVERT -j ACCEPT
+iptables -t mangle -I PREROUTING -p tcp -m socket -j DIVERT
```
obilodeau, line 161: I think I'll drop this. I'll test and drop if its not required.

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+# Disable return path filtering
+echo 0 > /proc/sys/net/ipv4/conf/default/rp_filter
+echo 0 > /proc/sys/net/ipv4/conf/all/rp_filter
+echo 0 > /proc/sys/net/ipv4/conf/enp0s3/rp_filter
+echo 0 > /proc/sys/net/ipv4/conf/enp0s8/rp_filter
+
+# Make interception work without an IP assigned to br0
+# WARNING: unclear if required on not
+echo 1 > /proc/sys/net/ipv4/conf/br0/route_localnet
+echo 1 > /proc/sys/net/ipv4/conf/enp0s3/route_localnet
```
obilodeau, line 154: I'll test and drop if not required

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+
+# WARNING: if you have other important firewall rules, make sure to test the impact of this
+modprobe br_netfilter
+echo 1 > /proc/sys/net/bridge/bridge-nf-call-iptables
+
+# Target DROP in brouting means out of L2 go route in L3
+ebtables -t broute -A BROUTING -i enp0s3 -p ipv4 --ip-dst $SERVER_IP --ip-proto tcp --ip-dport 3389 -j redirect --redirect-target DROP
+ebtables -t broute -A BROUTING -i enp0s8 -p ipv4 --ip-src $SERVER_IP --ip-proto tcp --ip-source-port 3389 -j redirect --redirect-target DROP
+# recent linux like debian buster
+#ebtables-legacy -t broute -A BROUTING -i enp0s8 -p ipv4 --ip-src $SERVER_IP --ip-proto tcp --ip-source-port 3389 -j redirect --redirect-target DROP
```
obilodeau, line 143: just point to the existence of ebtables-legacy instead of a whole copy of the statement

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+arp -i br0 -s 10.13.37.111 08:00:27:59:05:fe
+```
+
+Lastly, the MITM networking stack needs to know which traffic to route through the bridge, and this can complicate
+things significantly if a management interface is required. The simplest option is to have a route for the
+bridged subnet:
+
+```
+ip route add 10.13.37.0/24 dev br0
+```
```
obilodeau, line 206: I'm not convinced this is the correct explanation since a bridge just bridges the stuff through no matter what. I think it matters when we are putting packets (spoofed using `IP_TRANSPARENT`) in the kernel then the kernel knows where (at interface-level) to send them or what rules they need to go through (like mangle with pre-routing iface set to br0).

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+When using namespaces, all of the configuration for the MITM (from the previous sections) must be done while inside the
+namespace, and PyRDP must also be run from within the namespace. This creates a very clean logical split between the
+MITM host and PyRDP. You will need a third management interface for the host to remain reachable.
+
+It might also be a good idea to disable outbound ARP requests to prevent the host from accidentally sending ARP requests
+to the bridge:
+
+```bash
+arptables -A OUTPUT -o br0 -j DROP
+```
```
obilodeau, line 254: Killing ARP is incompatible with Option 1 (have an IP in that subnet) which is the one I would personally deploy if I were to do this somewhere. I think a re-org of namespace inside L2 approach and keeping the two options would be best.

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+
+# Useful References
```
obilodeau, line 256: add a namepsace reference

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+things significantly if a management interface is required. The simplest option is to have a route for the
+bridged subnet:
+
+```
+ip route add 10.13.37.0/24 dev br0
+```
+
+Other scenarios might require creativity with route configuration and are left to the reader's imagination.
+
+## Network Namespacing: Isolating MITM Interfaces
```
obilodeau, line 210: This approach is missing the default gateway, w/o it you can't reply to non-linklocal RDP connections

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+arp -i br0 -s 10.13.37.1 08:00:27:59:05:f0
+arp -i br0 -s 10.13.37.111 08:00:27:59:05:fe
+```
+
+Lastly, the MITM networking stack needs to know which traffic to route through the bridge, and this can complicate
+things significantly if a management interface is required. The simplest option is to have a route for the
+bridged subnet:
+
+```
+ip route add 10.13.37.0/24 dev br0
```
obilodeau, line 205: Without the default gateway you can't reply to non-local RDP connections. 192.168.0.201 isn't for me, I don't have a route, I don't have a default route, good bye.

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+### Option 1: Add an Interface with a valid IP in the networks of interests on the Bridge
+
+This is important to allow ARP resolution to work properly (when the MITM unavoidably needs to know who has the
+server/client IP)
+
+In the example setup, doing something like below is enough to make everything work because when the MITM needs the L2
+address to reply to, it has the means to look it up by itself.
+
+```
+ip addr add 10.13.37.11/24 dev br0
```
obilodeau, line 175: Without the default gateway you can't reply to non-local RDP connections. 192.168.0.201 isn't for me, I don't have a route, I don't have a default route, good bye.

docs/transparent-proxy.md
```diff
@@ -0,0 +1,268 @@
...
+
+This document provides example network configurations for transparent
+proxying (TPROXY) Its purpose is to provide a starting point for
+people with specific transparent proxy needs. Network configuration
+can and will vary significantly depending on the environment in which
+PyRDP is deployed. It is likely that the examples below will not work
+as-is and require modifications and testing during deployment.
+
+
+## Basic L3 TPROXY
```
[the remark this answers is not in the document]
alxbl: It is.

docs/transparent-proxy.md
alxbl: It wasn't required on my setup, but do confirm on your end as well.

docs/transparent-proxy.md
alxbl: Agreed.

docs/transparent-proxy.md
alxbl: Yeah. This should be documented.

alxbl:
I assume this means you'll be taking over the PR and committing those changes you mentioned. I'll let you take care of it, otherwise let me know.

docs/transparent-proxy.md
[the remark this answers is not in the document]
obilodeau: Fixed

docs/transparent-proxy.md
obilodeau: Removed

docs/transparent-proxy.md
obilodeau: Dropped it

docs/transparent-proxy.md
obilodeau: Mentioned at the end of option 2.

docs/transparent-proxy.md
obilodeau: Done

docs/transparent-proxy.md
obilodeau: Fixed

docs/transparent-proxy.md
obilodeau: Fixed

docs/transparent-proxy.md
obilodeau: Fixed

alxbl:
LGTM!

Issue #51, opened 2018-12-10
Add class for client scripting with asyncio


xshill:
It would be nice to have a class that simplifies client scripting. The goal is to make it easy to write simple scripts for interacting with an RDP server. I think an API like this would be good:

```
async def main():
    rdp = RDPClient()
    await rdp.connect("127.0.0.1", "test", "pass", 3000)
    await rdp.setMouse(200, 300) # Pretend this is an icon for a text file
    await rdp.doubleClick()
    await rdp.type("Hello World!")
    await rdp.sendSequence("Ctrl+S")
    await rdp.sendSequence("Alt+F4")
```

The RDPClient class would use the async/await syntax (which works with Twisted with asyncio integration). It would define the high-level interactions one can do with RDP (e.g: move the mouse, send key presses, send mouse clicks, etc.). In addition to these methods, it could be useful if the client exposed the screen as an array of pixels so we can do something with the image.

PR #206, opened 2020-04-08
Improve PyRDP High Level Documentation

### LINKED ISSUE #62: Add documentation on using PyRDP as a library
We should have documentation that explains how people can use PyRDP to create their own RDP-related tools. Here are a list of subjects that should be covered:

- The relations between PDU, Parser, Layer and Observer classes and their purpose
- A brief description of every package in pyrdp
- The connection sequence (more briefly than MSDN) along with a commented example script
- A brief explanation of MCS
- An explanation of the differences between all the security protocols that can be used (NLA protocols, TLS and RC4)
- Things that we circumvent in our MITM and that might need to be circumvented by other tools (e.g: NLA)

alxbl:
This branch adds high-level documentation for PyRDP's architecture, modules, and extension points.

It should serve as both a reference for people wishing to contribute to PyRDP, and as a tour of the libraries for people wishing to script using PyRDP.

It is a work in progress right now as there are several code cleanup opportunities that came up while writing this documentation. I'd like to implement those as well before landing this. In the process it might be worthwhile to investigate #51, but that seems like it would potentially be a large effort.

----

Closes #62

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,9 +1,131 @@
 # The PyRDP library
 Our RDP implementation was designed so it could be used for tools other than our MITM.
 
-This section is still under construction. It will explain how to use the classes in PyRDP to achieve your goals.
-In the meantime, we advise you to look at the following resources:
+## Quick Resources
 
-- Our documentation on [layers](layers.md).
-- [The RDP connection sequence](https://msdn.microsoft.com/en-us/library/cc240452.aspx).
-- Our MITM implementation (`pyrdp.mitm.client` and `pyrdp.mitm.server`).
+- The PyRDP [Networking Layers Architecture][layers]
+- [MS-RDPBCGR][MS-RDPBCGR]: Basic Connectivity and Graphics Remoting
+- [MS-RDPEGDI][MS-RDPEGDI]: The Graphics Device Interface Extensions
+- [RDP Connection Sequence][connection]
+
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,9 +1,131 @@
...
+- BitmapParser
+
+
+The **player** module contains code and logic for the Player component
+to render an actual RDP session into various forms. The most common
+form is to render to the player window, but other types of rendering
+are possible (text-only, for example) and can be implemented here.
+
+
+The **recording** module offers capabilities to export an RDP session
```
Res260, line 110: Not just disk, over the network as well

alxbl: The "sending over the network" part is not happening in the recording module. The recorder must be given a PDU and it passes it into its player layers blindly without caring where the data ends up, which may or may not include network layers, but is currently exclusively a FileLayer in the codebase (AFAICT.)

In fact, while I was reading the recorder code to reply to your comment, I realized that the whole recorder could have just been be an extra layer that composes/subclasses the PlayerLayer. It's too late to refactor it and not really worth it, though. The actual recording is done by the `FileLayer`. I still think that for documentation clarity, the main use case of the module should be documented. If you don't agree, feel free to provide the description you would like to see for the recording module, and I'll update the text to include it.

alxbl:
In the interest of landing vNext sooner rather than later, I am going to forego the refactors that I had intended to do. We can always revisit this later if/when the architecture changes.

PR #207, opened 2020-04-08
fix: Buffer replays until connection is established. Fixes #169

### LINKED ISSUE #169: 0/18 byte files in replay folder
Is there a way to stop it from making replay files and the player tabs on every scan? It makes it rather difficult to sort through.

alxbl:
Fixes #169

### DIFF
--- a/pyrdp/recording/recorder.py
+++ b/pyrdp/recording/recorder.py
@@ -4,6 +4,7 @@
 # Licensed under the GPLv3 or later.
 #
 
+from io import BufferedIOBase
 from pathlib import Path
 from typing import Dict, List, Optional, Union
 
@@ -95,7 +96,7 @@ class FileLayer(LayerChainItem):
         self.pending = b''
 
         self.filename = fileName
-        self.fd = None
+        self.fd: BufferedIOBase = None
 
     def sendBytes(self, data: bytes):
         """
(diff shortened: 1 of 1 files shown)


pyrdp/recording/recorder.py
```diff
@@ -86,17 +86,30 @@ class FileLayer(LayerChainItem):
...
+
     def __init__(self, fileName: Union[str, Path]):
         super().__init__()
-        self.file_descriptor = open(str(fileName), "wb")
+
+        # Buffer data until we have enough bytes for a meaningful replay.
+        self.pending = b''
+
+        self.filename = fileName
+        self.fd = None
```
Res260, line 98: I wouldve kept `fileDescriptor` because

>Explicit is better than implicit.

and
>Readability counts.

alxbl: I would argue that `fd` is very explicit and readable as the linux API and manpages are always referring to `fd`s in all of its signatures for files, and it's very common to have the constructor as `with open(x) as f:` or `as fd:` The class also fits in a screenful...

Res260: Fair enough. However, to align with the rest of the codebase, I'd put the variable type (`io.TextIOWrapper`) because the variable is instanciated to None, which removes completion capability.

alxbl: I'm pretty sure `TextIOWrapper` is not the right type, since this is a binary file. I've added type hinting for BufferedIOBase, which exports the right `write()` signature.

Res260: My bad, I looked to quickly. BufferedIOBase looks good indeed!

obilodeau:
Good

PR #208, opened 2020-04-09
Scale pyrdp-player replays with window height

### LINKED ISSUE #101: Player should support window scaling
Hi, Some suggestions for the player.

1. Replay window scaling would be very helpful.
2. ~notify2 is totally incompatible with ms-windows, I think. I use the player with import notify2 and its calls commented out (but otherwise it works well, thanks).~ done in #129

Res260:
image

image

Works ONLY on replays, not on the live player tab (because I was too scared to commit to testing the whole session takeover feature to be bug free).

~Right now, I will keep it as a draft because it is the default option, but I want to have a checkbox where we can choose to scale or not.~

EDIT: Checkbox has been implemented, images updated

Also maximizes pyrdp-player upon launch

Fixes #101

### DIFF
--- a/pyrdp/player/MainWindow.py
+++ b/pyrdp/player/MainWindow.py
@@ -3,10 +3,13 @@
 # Copyright (C) 2018, 2019 GoSecure Inc.
 # Licensed under the GPLv3 or later.
 #
+from typing import List
 
 from PySide2.QtCore import Qt, Signal
-from PySide2.QtWidgets import QAction, QFileDialog, QMainWindow, QTabWidget, QInputDialog
+from PySide2.QtGui import QResizeEvent
+from PySide2.QtWidgets import QAction, QFileDialog, QInputDialog, QMainWindow, QTabWidget
 
+from pyrdp.player import BaseTab
 from pyrdp.player.LiveWindow import LiveWindow
 from pyrdp.player.ReplayWindow import ReplayWindow
 
@@ -32,13 +35,14 @@ class MainWindow(QMainWindow):
             "closeTabOnCtrlW": True     # Allow user to toggle Ctrl+W passthrough.
         }
 
-        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options)
-        self.replayWindow = ReplayWindow(self.options)
+        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options, mainWindow=self)
+        self.replayWindow = ReplayWindow(self.options, mainWindow=self)
         self.tabManager = QTabWidget()
         self.tabManager.addTab(self.liveWindow, "Live connections")
         self.tabManager.addTab(self.replayWindow, "Replays")
         self.setCentralWidget(self.tabManager)
         self.updateCountSignal.connect(self.updateTabConnectionCount)
+        self.resizeObservers: List[BaseTab] = []
 
         # File menu
         openAction = QAction("Open...", self)
@@ -110,7 +114,6 @@ class MainWindow(QMainWindow):
             for fileName in fileNames:
                 self.replayWindow.openFile(fileName)
 
-
     def sendKeySequence(self, keys: [Qt.Key]):
         if self.tabManager.currentWidget() is self.liveWindow:
             self.liveWindow.sendKeySequence(keys)
@@ -138,3 +141,12 @@ class MainWindow(QMainWindow):
         """
 
         self.tabManager.setTabText(0, "Live connections (%d)" % self.liveWindow.count())
+
+    def addObserver(self, observer: BaseTab):
+        self.resizeObservers.append(observer)
+        observer.mainWindowResized(self.width(), self.height())
+
+    def resizeEvent(self, event: QResizeEvent):
+        for observer in self.resizeObservers:
+            observer.mainWindowResized(event.size().width(), event.size().height())
+        super().resizeEvent(event)
--- a/pyrdp/player/MainWindow.py
+++ b/pyrdp/player/MainWindow.py
@@ -1,6 +1,6 @@
 #
 # This file is part of the PyRDP project.
-# Copyright (C) 2018, 2019 GoSecure Inc.
+# Copyright (C) 2018-2020 GoSecure Inc.
 # Licensed under the GPLv3 or later.
 #
 from typing import List
--- a/pyrdp/player/MainWindow.py
+++ b/pyrdp/player/MainWindow.py
@@ -35,7 +35,7 @@ class MainWindow(QMainWindow):
         }
 
         self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options, parent=self)
-        self.replayWindow = ReplayWindow(self.options, mainWindow=self, parent=self)
+        self.replayWindow = ReplayWindow(self.options, parent=self)
         self.tabManager = QTabWidget()
         self.tabManager.addTab(self.liveWindow, "Live connections")
         self.tabManager.addTab(self.replayWindow, "Replays")
--- a/pyrdp/player/MainWindow.py
+++ b/pyrdp/player/MainWindow.py
@@ -6,7 +6,6 @@
 from typing import List
 
 from PySide2.QtCore import Qt, Signal
-from PySide2.QtGui import QResizeEvent
 from PySide2.QtWidgets import QAction, QFileDialog, QInputDialog, QMainWindow, QTabWidget
 
 from pyrdp.player import BaseTab
@@ -35,8 +34,8 @@ class MainWindow(QMainWindow):
             "closeTabOnCtrlW": True     # Allow user to toggle Ctrl+W passthrough.
         }
 
-        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options)
-        self.replayWindow = ReplayWindow(self.options, mainWindow=self)
+        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options, parent=self)
+        self.replayWindow = ReplayWindow(self.options, mainWindow=self, parent=self)
         self.tabManager = QTabWidget()
         self.tabManager.addTab(self.liveWindow, "Live connections")
         self.tabManager.addTab(self.replayWindow, "Replays")
@@ -141,12 +140,3 @@ class MainWindow(QMainWindow):
         """
 
         self.tabManager.setTabText(0, "Live connections (%d)" % self.liveWindow.count())
-
-    def addObserver(self, observer: BaseTab):
-        self.resizeObservers.append(observer)
-        observer.mainWindowResized(self.width(), self.height())
-
-    def resizeEvent(self, event: QResizeEvent):
-        for observer in self.resizeObservers:
-            observer.mainWindowResized(event.size().width(), event.size().height())
-        super().resizeEvent(event)
--- a/pyrdp/player/MainWindow.py
+++ b/pyrdp/player/MainWindow.py
@@ -35,7 +35,7 @@ class MainWindow(QMainWindow):
             "closeTabOnCtrlW": True     # Allow user to toggle Ctrl+W passthrough.
         }
 
-        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options, mainWindow=self)
+        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options)
         self.replayWindow = ReplayWindow(self.options, mainWindow=self)
         self.tabManager = QTabWidget()
         self.tabManager.addTab(self.liveWindow, "Live connections")
--- a/pyrdp/player/ReplayTab.py
+++ b/pyrdp/player/ReplayTab.py
@@ -108,4 +108,3 @@ class ReplayTab(BaseTab):
         """
         newScale = (self.scrollViewer.height() - self.scrollViewer.horizontalScrollBar().height()) / self.widget.sessionHeight
         self.widget.scale(newScale)
-        self.thread.parentResized()
--- a/pyrdp/player/ReplayTab.py
+++ b/pyrdp/player/ReplayTab.py
@@ -1,6 +1,6 @@
 #
 # This file is part of the PyRDP project.
-# Copyright (C) 2019 GoSecure Inc.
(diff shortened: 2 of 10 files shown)


pyrdp/player/ReplayTab.py
```diff
@@ -88,3 +89,13 @@ def clear(self):
...
         self.thread.wait()
+
+    def mainWindowResized(self, width: int, height: int):
+        """
+        Called when the main PyRDP window is resized to allow to scale the current
+        RDP session being displayed.
+        :param width: The new width of the main window
+        :param height: The new height of the main window
+        """
+        self.widget.scale((height - self.text.height() - 200) / self.widget.height)
```
alxbl, line 100: What is this `200` for? I assume it's the height of the settings pane at the top? If that height is hardcoded, it should probably be a constant to avoid rescaling bugs if we ever change the GUI. In fact I would think that Qt has a way for you to query that information, so that the UI is responsive.

pyrdp/player/ReplayWindow.py
```diff
@@ -17,15 +18,17 @@ class ReplayWindow(BaseWindow):
     Class for managing replay tabs.
     """
 
-    def __init__(self, options: Dict[str, object], parent: QWidget = None):
-        super().__init__(options, parent)
+    def __init__(self, options: Dict[str, object], mainWindow: MainWindow, parent: QWidget = None):
```
alxbl, line 21: Can't we just use the parent QWidget argument here? isn't that the whole point of Qt?  the `ReplayTab` is logically a child of the QWindow. I would test that as it would keep the constructor the same and be more consistent with Qt You can still cache the main window to avoid having to call `self.getParent()`.

Res260:
> This is a very cool PR. Thanks for taking the time to work on this!
> 
> My only critique is that you seem to be ignoring the concept of parent widgets in Qt. I have a feeling that if we are properly setting the widget chain (MainWindow -> ReplayWindow -> ReplayTab), then the `onResize` events from Qt would propagate throughout the widget tree and avoid all of the need for this observable code and manual event dispatching. I am not sure, but I know a lot of GUI frameworks work like that, so if you can test it, it would be nice and greatly simplify the code.
> 
> If having parent widgets introduces other bugs, I am willing to accept this implementation first and iron out the bugs/refactor to be more idiomatic Qt later down the road.

I'll try and test this today and report back. I didnt think the resize events would propagate, but havent tested

pyrdp/player/ReplayTab.py
Res260: With my current implementation it was hard to query this data, I'll see if it's easier with your suggestion

pyrdp/player/ReplayWindow.py
Res260: I indeed found a more intelligent way to handle this, see the latest commit! thanks

pyrdp/player/ReplayTab.py
Res260: not needed anymore

Res260:
It should be a better implementation now. It doesnt use the parent concept, but it is way cleaner still.

pyrdp/player/ReplayWindow.py
```diff
@@ -27,9 +28,13 @@ def openFile(self, fileName: str, autoplay: bool = False):
...
         """
-        tab = ReplayTab(fileName)
-        self.mainWindow.addObserver(tab)
+        tab = ReplayTab(fileName, parent=self)
         self.addTab(tab, fileName)
         self.log.debug("Loading replay file %(arg1)s", {"arg1": fileName})
         if autoplay:
             tab.play()
+
+    def resizeEvent(self, event: QResizeEvent):
```
alxbl, line 37: I assume this means that the resize event was not being propagated?

pyrdp/player/ReplayThread.py
```diff
@@ -107,4 +107,11 @@ def setSpeed(self, speed: float):
...
     def close(self):
-        self.queue.put(ReplayThreadEvent.EXIT)
\ No newline at end of file
+        self.queue.put(ReplayThreadEvent.EXIT)
+
+    def parentResized(self):
+        """
+        Seeks to current time to allow to rerender the image on the resized widget.
+        """
+        self.lastSeekTime = self.timer.getElapsedTime()
```
alxbl, line 116: Um, I don't think seeking should be necessary to support scaling.

This is likely an indicator that the logic is not being handled in the right place. We have a `QRemoteDesktop` widget that always has the latest rendered frame in a `QImage` buffer. The easiest way to handle the scaling would be to propagate the resize event all the way to it, and have it keep the original size buffer and paint the scaled buffer whenever a window repaint is requested.

Does what I'm saying make sense? I have google a little bit and found that people are successfully propagating the resize event down their widget tree, so this doesn't sound impossible.

In the worst case scenario, I don't mind manual propagation all the way to the QRemoteDesktop widget, but I think that that's where it should be handled.

pyrdp/player/ReplayWindow.py
Res260: Well resizing the main window does trigger the resizeEvent of the ReplayWindow because of the layout type used, but it does not trigger the QRemoteDesktop resizeEvent indeed

pyrdp/player/ReplayThread.py
alxbl: Okay, I've read a bit more. It seems like events will bubble up, but not drill down into the tree, so your code handling resize by calling a resize handler on the children is unavoidable AFAICT. But the rest of this comment still stands.

pyrdp/player/ReplayWindow.py
```diff
@@ -17,16 +19,22 @@ class ReplayWindow(BaseWindow):
...
+        super().resizeEvent(event)
+        for i in range(self.count()):
+            self.widget(i).parentResized(event.size().width(), event.size().height())
```
alxbl, line 40: You could just pass the event here, since all of these are widgets anyway.

pyrdp/player/ReplayThread.py
Res260: This is already the case. QRemoteDesktop's `_buffer` contains the unscaled image, which is then rescaled at render time.

This is a good catch. Your comment made me realise that I only did this because the previous implementation created a new blank image upon changing the scaling, but I can simply copy the previous buffer onto the new one, which removes the need to seek.

pyrdp/player/ReplayWindow.py
Res260: I don't even actually use these arguments, I put them because I thought it would be a bad api for the children if i didnt put them. Should I remove them?

alxbl: If the purpose of the code is to drill the event down the widget tree for children to be notified of a parent's resize, it makes sense to have the event as an argument. 

Otherwise don't make the code generic and just make a method that works directly on the `ReplayTab`s that's called `resize` or something.

Res260: i went with the first approach

pyrdp/player/MainWindow.py
```diff
@@ -32,13 +34,14 @@ def __init__(self, bind_address: str, port: int, filesToRead: [str]):
             "closeTabOnCtrlW": True     # Allow user to toggle Ctrl+W passthrough.
         }
 
-        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options)
-        self.replayWindow = ReplayWindow(self.options)
+        self.liveWindow = LiveWindow(bind_address, port, self.updateCountSignal, self.options, parent=self)
+        self.replayWindow = ReplayWindow(self.options, mainWindow=self, parent=self)
```
alxbl, line 38: So I feel like I'm commenting this for the third time: You realize you are passing `self` twice for 2 different parameters, right? The whole `mainWindow` parameter is redundant because MainWindow **is** the parent of the ReplayWindow.

pyrdp/ui/qt.py
```diff
@@ -174,26 +183,40 @@ def setMousePosition(self, x: int, y: int):
...
+        self.scaleToWindow = status > 0
+
     def resize(self, width: int, height: int):
         """
         Resize widget
         :param width: new width of the widget
         :param height: new height of the widget
         """
-        self._buffer = QImage(width, height, QImage.Format_RGB32)
+        self._buffer = self._buffer.scaled(width, height)
```
alxbl, line 203: You shouldn't resize the rendering buffer. This will mess with PyRDP's rendering pipeline. You need to keep the original buffer full size and only scale it when it's time to paint the widget. I don't think it's going to be a performance issue, but if it is, we might need to have a debounce if Qt spams repaint events. in practice we'll be minimum-bound by the framerate, though.

pyrdp/ui/qt.py
```diff
@@ -174,26 +183,40 @@ def setMousePosition(self, x: int, y: int):
         self.mouseY = y
         self.update()
 
+    def scale(self, scale):
+        """
+        Rescale the current widget to a percentage of the height of the RDP session.
+        :param scale: Ex: 0.5 for 50% height and 50% width.
+        """
+        self.scaleValue = scale
+        self.resize(self.sessionWidth, self.sessionHeight)
```
alxbl, line 192: No.

The rendering surface that the drawing pipeline use must remain the full resolution of RDP, since all bitmap coords and offset are based on that resolution. Scaling should only happen at the very last moment before painting.

pyrdp/ui/qt.py
```diff
@@ -174,26 +183,40 @@ def setMousePosition(self, x: int, y: int):
...
 
     def paintEvent(self, e: QEvent):
         """
         Call when Qt renderer engine estimate that is needed
         :param e: the event
         """
+        scaleValue = self.scaleValue if self.scaleToWindow else 1
         qp = QPainter(self)
-        qp.drawImage(0, 0, self._buffer)
+        qp.drawImage(0, 0, self._buffer.scaled(self.sessionWidth * scaleValue, self.sessionHeight * scaleValue, aspectMode=Qt.KeepAspectRatio))
```
alxbl, line 215: This is correct. This is where the scaling should happen. Also, FYI: a scaling factor is called a `ratio` or factor.

pyrdp/player/MainWindow.py
Res260: god damn I keep thinking i removed everything from the previous implementation. will remove

pyrdp/ui/qt.py
Res260: This is already a case. You're confusing `resize` and `scale`. `resize` already existed and was used to update the buffer size according to the screen size **when we receive clientData**. _buffer is only truely resized once, when clientData is parsed.

Hovewer, it is true that this is currently confusing and made me realise that `scale()` doesnt have to call `resize()` anymore, so I removed the call from `scale()`. I also clarified the docstring so its purpose is clearer.

pyrdp/ui/qt.py
Res260: see last comment

pyrdp/ui/qt.py
Res260: right I'll rename

pyrdp/ui/qt.py
```diff
@@ -134,12 +136,22 @@ def __init__(self, width: int, height: int, parent: QWidget = None):
...
+
+        self.scaleToWindow = False
+
+        self._buffer: QImage = QImage(width, height, QImage.Format_ARGB32_Premultiplied)
+
         # Set correct size
         self.resize(width, height)
         # Bind mouse event
         self.setMouseTracking(True)
         # Buffer image
```
alxbl, line 154: If you move the definition, move the comment too.

pyrdp/player/MainWindow.py
```diff
@@ -3,10 +3,12 @@
 # Copyright (C) 2018, 2019 GoSecure Inc.
```
obilodeau, line 3:
```suggestion
# Copyright (C) 2018-2020 GoSecure Inc.
```

You can call me the copyright bumper artist. :sunglasses: ... I know, it's pathetic. :sweat_smile:

pyrdp/player/ReplayTab.py
```diff
@@ -3,7 +3,7 @@
 # Copyright (C) 2019 GoSecure Inc.
```
obilodeau, line 3:
```suggestion
# Copyright (C) 2019, 2020 GoSecure Inc.
```

pyrdp/player/ReplayThread.py
```diff
@@ -44,41 +44,41 @@ def __init__(self, replay: Replay):
         self.lastSeekTime = 0
         self.requestedSpeed = 1
         self.replay = replay
+        self.timer = Timer()
```
obilodeau, line 47: Were these changes really necessary?

pyrdp/ui/qt.py
```diff
@@ -24,10 +24,10 @@
 QRemoteDesktop is a widget use for render in rdpy
 """
 
+import rle
 from io import BytesIO
 
-import rle
```
obilodeau, line 29: FYI: `rle` is not an stdlib module. Your import organizer is broken and you shouldn't have made that change.

pyrdp/ui/qt.py
```diff
@@ -151,7 +163,6 @@ def runOnMainThread(self, target: callable):
     @property
     def screen(self):
         return self._buffer
-
```
obilodeau, line 154: Not sure what this is doing here given your taste to reindent and add new lines to things :wink:

pyrdp/player/ReplayThread.py
Res260: It was in my previous implementation. I forgot to completely revert this change after I changed the implementation, so no they were not necessary.

pyrdp/ui/qt.py
Res260: PEP8 :)

obilodeau: pep8-wat

No empty lines between a property and a method? This is PEP8?

Res260: I missed this one :(

Res260: wait

Res260: i'm now ashamed of myself

obilodeau: Don't be

Res260: haha it was mostly a joke

PR #212, opened 2020-04-15
Add support for Ehanced Security Direct Connection

alxbl:
This is a DRAFT pull request that modifies PyRDP to support the direct connection approach outlined in [1.3.1.2 Security-Enhanced Connection Sequence](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/49e2791f-72d0-4229-92b7-8f32c48a7206).

I think it still needs some work because it relies on private Twisted APIs (Unavoidable?) and propagates the MITM config into non-MITM modules, which feels unclean.

I would like to also refactor the MITM Config to split the MITM specific settings and the Core settings. This could also plug into the `.ini` configuration files while we're playing in that code area.

I'm opening the PR now in case someone has time (and wants) to work on it.

### DIFF
--- a/pyrdp/layer/tcp.py
+++ b/pyrdp/layer/tcp.py
@@ -11,10 +11,15 @@ from binascii import hexlify
 from twisted.internet.protocol import connectionDone, Protocol
 
 from pyrdp.core import ObservedBy
+from pyrdp.core.ssl import ServerTLSContext
 from pyrdp.layer.layer import IntermediateLayer, LayerObserver
 from pyrdp.logging import LOGGER_NAMES, getSSLLogger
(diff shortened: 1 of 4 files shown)


pyrdp/layer/tcp.py
```diff
@@ -46,7 +45,7 @@ class TwistedTCPLayer(IntermediateLayer, Protocol):
     TCP observers are notified when a connection is made.
     """
 
-    def __init__(self, config: MITMConfig):
```
xshill: Pretty sure you can just put quotes around the class name to avoid importing it.

PR #213, opened 2020-04-15
Minor tweaks to the window scaling code

obilodeau:
Two main changes

### Scrollbar only displayed if replay is not scaled (the default)

It's fixed. Dispatched events needed to be re-ordered because resize was called before QRemoteDesktop's state was updated and I was querying it.

### Few pixels were missing when scaled

Before change:
pyrdp-pixels-grugés

After change:
pyrdp-pixels-ok

The difference between using the viewport() and the height minus scrollbar height is 2 pixels but that was enough to be perceptible.

### DIFF
--- a/pyrdp/ui/qt.py
+++ b/pyrdp/ui/qt.py
@@ -194,6 +194,18 @@ class QRemoteDesktop(QWidget):
         :param scale: Ex: 0.5 for 50% height and 50% width.
         """
         self.ratio = scale
+        self._updateWidgetSize()
+
+    def _updateWidgetSize(self):
+        """
+        Size the widget according to if we are scaled or not
+        """
+        if self.scaleToWindow:
+            # resize widget to parent size: will hide scrollbars
+            super().resize(self.parent().size())
+        else:
+            # resize widget to session size: will show scrollbars if required
+            super().resize(self.sessionWidth, self.sessionHeight)
 
     def setScaleToWindow(self, status):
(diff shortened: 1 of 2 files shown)


pyrdp/ui/qt.py
```diff
@@ -166,6 +166,7 @@ def runOnMainThread(self, target: callable):
     @property
     def screen(self):
         return self._buffer
+
```
Res260, line 169: <3

pyrdp/ui/qt.py
```diff
@@ -194,6 +195,12 @@ def scale(self, scale):
         :param scale: Ex: 0.5 for 50% height and 50% width.
         """
         self.ratio = scale
+        if not self.scaleToWindow:
+            # resize widget to session size
+            super().resize(self.sessionWidth, self.sessionHeight)
```
Res260, line 200: wouldn't it be more readable using `if condition x else y`  instead of `if not condition y else x`?

obilodeau: Good point

obilodeau: fixed

obilodeau:
Managed to fix it. Give it a try @Res260.

Res260:
Works well!

obilodeau:
Now that review has happened, I `rebased -i` and squashed fixup commits together. Ready to merge when CI finishes.

PR #214, opened 2020-04-16
Fix CI not completely running on windows

Res260:
~The latest CI failure made me wonder why it failed on ubuntu and not on windows even though I tested it to fail on my windows machine as well. I think I spotted the bug (currently testing)~

## The problem:
The CI used powershell with this to run the desired command: 
image

For some reason, running directly the python script seemed to act as a NOOP and exited with code 0, making the CI step a success even though the CI didn't actually execute the script.

## The fix:
call `python` before the script so it actually runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
 
       - name: pyrdp-mitm.py initialization test
         working-directory: ./
-        run: python bin/pyrdp-mitm.py --test localhost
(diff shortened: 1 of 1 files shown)


obilodeau:
The actual reason it doesn't work is because Windows doesn't execute script (non-PE) files on it's own like Unix operating systems do. Read about the shebang line. On Windows this happens with file associations which are based on extensions.

Adding the interpreter explicitly on Windows is excellent!

PR #215, opened 2020-04-16
CI installs the full package (GUI+AV) instead of slim

obilodeau:
It will catch more errors as they are introduced. For example, adding the ffmpeg dependency will fail a build on Ubuntu 18.04. The docker build caught it but not the Ubuntu install.

I think doing it on Windows would be good too. Hopefully, we will be able to turn it on.

I expect this PR to be rocky with failures and finish with a squash merge.


obilodeau:
It passes on the first go? _/me pinches itself_

PR #218, opened 2020-04-17
Install openssl in the docker image

### LINKED ISSUE #216: Certificate generation failed, starting docker container
When attempting to run the docker container. Not sure if it's something I'm doing wrong. 

docker run -p 3389:3389 gosecure/pyrdp pyrdp-mitm.py 10.x.x.x

[2020-04-17 01:41:50,001] - INFO - GLOBAL - pyrdp.mitm - Generating a private key and certificate for SSL connections
[2020-04-17 01:41:50,012] - ERROR - GLOBAL - pyrdp.mitm - Generation failed. Please provide the private key and certificate with -k and -c
[2020-04-17 01:41:50,013] - ERROR - GLOBAL - pyrdp - An error occurred when creating the server TLS context. There may be a problem with your private key or certificate (e.g: signature algorithm too weak). Here is the exception: [('system library', 'fopen', 'No such file or directory'), ('BIO routines', 'file_ctrl', 'system lib'), ('SSL routines', 'SSL_CTX_use_certificate_file', 'system lib')]

Res260:
OpenSSL was not installed in the runtime image, so the MITM probably wasnt working at all. I'm not sure how that could be tested in the CI.

Fixes #216

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ FROM ubuntu:18.04 AS docker-image
 
 # Install runtime dependencies except pre-built venv
 RUN apt-get update && apt-get install -y --no-install-recommends python3 \
+        openssl \
         # GUI and notifications stuff
         libgl1-mesa-glx \
         notify-osd dbus-x11 libxkbcommon-x11-0 \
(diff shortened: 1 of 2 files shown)


Res260:
Right I forgot the slim image. Should be good now!

PR #220, opened 2020-04-18
Add a CI badge in the readme

Res260:
Just discovered this felt the need to add it

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Continuous Integration
+name: Build
 
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


Res260:
just looked at the diff its vscode's fault :(

Res260:
@alxbl Yeah we would need to change the `name` in CI.yml to `Build` and then change the url of the badge.

alxbl:
If it's not too much a hassle, I'd rather have the badge be called "Build" since that's what most people expect and what most projects use.

Res260:
done @alxbl

Res260:
(should be squashed)

PR #227, opened 2020-05-21
Enable GDI graphics pipeline by default.

### LINKED ISSUE #209: GDI should be enabled by default
As discussed in #187 which introduces GDI support: We will keep the option disabled by default for now but we will enable it by default for our next major release which will be 1.0.

This is a tracker attached to our next release milestone so we don't forget.

alxbl:
Closes #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ research use cases in mind.
     + [Other MITM arguments](#other-mitm-arguments)
       - [--no-downgrade](#--no-downgrade)
       - [--transparent](#--transparent)
-      - [`--gdi`: Accelerated Graphics Pipeline](#--gdi-accelerated-graphics-pipeline)
+      - [`--no-gdi`: Disable Accelerated Graphics Pipeline](#--no-gdi-disable-accelerated-graphics-pipeline)
   * [Using the PyRDP Player](#using-the-pyrdp-player)
     + [Playing a replay file](#playing-a-replay-file)
     + [Listening for live connections](#listening-for-live-connections)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -368,11 +368,11 @@ a gateway between clients and the server and sees all traffic.
 **NOTE**: This requires root privileges, only works on Linux and requires manual firewall configuration to ensure 
 that traffic is routed properly.
 
-##### `--gdi`: Accelerated Graphics Pipeline
+##### `--no-gdi`: Disable Accelerated Graphics Pipeline
 
-Tells the MITM to allow clients to use [Graphics Device Interface Acceleration][gdi] Extensions to stream
-drawing orders instead of raw bitmaps. The advantage of this mode is a significant reduction in required bandwidth
-for high resolution connections.
+By default, PyRDP uses the most recent supported graphics pipeline. This switch tells the MITM to not use the [Graphics
```
Res260, line 373: I think this might be confusing, because AFAIK GFX is the most recent graphics pipeline used in RDP. Even though PyRDP doesnt disable GFX support (IIRC), but it is not used when going through PyRDP (tested with win10 <--> pyrdp <--> win10). Further testing might be required, it has been a while since I did those tests.

alxbl: "Most recent supported". We only support bitmap and GDI.

I can change to "most recent pipeline supported by PyRDP" if you think that's clearer.

Res260: That would repeat "pyrdp". I think it might bug me because it's the server that chooses how it handles graphics, not PyRDP (which only downgrades to supported protocols). I'm not sure of a better way to phrase it, if I think about one I'll open a separate pr!

PR #228, opened 2020-05-25
Add Support for Carving Clipboard Files

### LINKED ISSUE #100: Collect clipboard file transfer
Hi,

Is it possible to store under `pyrdp_output/files` files transfered using clipboard?

Thanks.

alxbl:
This PR adds support for parsing file transfer clipboard PDUs and capabilities for the ClipboardMITM to dump files to disk.

**What it does**

- Tracks the clipboard file list
- Tracks clipboard transfers
- Dumps files to disk as they are transferred

**What's left to do**
- [x] Respect the configuration for file output
- [x] Implement transfer timeouts and cancellations to avoid dangling file handles.
- [x] Documentation

**Not Supported**
- The remote clipboard could in theory transfer a file's chunks out of order...
-----
Closes #100


alxbl:
Fixed merge conflicts. I'm going to try to get a working timeout prototype this week for abandoned transfers, or at the very least a clean up mechanism when the connection is lost so that we don't leak file handles. In practice this is unlikely to happen for non-malicious clients, but I'd rather be safe.

PR #229, opened 2020-06-05
feat: Added CredSSP support when private key is known

### LINKED ISSUE #134: NLA support when we control the destination server
How to implement client in connection honeypot service, let client input username, password

### LINKED ISSUE #154: Document the extraction of trusted certificate from Windows destination RDP to avoid untrusted certs errors
I have access to the rdp server and need to clone / export a trusted certificate
I tried:
- pyrdp-clonecert.py
- Export certificate from server without key.
- Export certificate and key from server using mimikatz
- Export the certificate from the mitm server and run pyrdp-mitm with this certificate and key.

The certificate has always been untrusted!
How do I clone a certificate to be trusted?

alxbl:
This pull request adds a `--auth` switch that allows to selectively enable the authentication mechanisms that PyRDP will allow through. For now these only include `tls` and `ssp` with some strict limitations for CredSSP.

It also finally documents how to extract the private key of RDP servers so that we can close #154 
Also resolves #134.

## Details

- Does not support the actual negotiation of CredSSP, which would require a lot of code and knowledge of the NTLM hash or the target's credentials.
- Does not support the [CredSSPY][0] attack that was just released.
- In order for the connection to be MITMed, it requires the private key of the server

In practice, this means that the only purpose of this feature is to allow honeypots to expose CredSSP. Note that usage of CredSSP means credential stuffing is not possible since a secret key is derived from the NTLM hash of the credentials.

[0]: https://github.com/croustibaie/CredSSPY/

## Future Work

This pull-request also paves the way towards supporting other authentication techniques such as Early User Authorization (EUA), but that is not yet implemented.

### DIFF
--- a/docs/cert-extraction.md
+++ b/docs/cert-extraction.md
@@ -4,7 +4,7 @@ This procedure is useful when running honeypots to support CredSSP (using `--aut
 It requires Administrative privileges on the target server and the use of Mimikatz, so it assumes that you are able to deactivate the Anti-Virus on the target server.
 
 
-> **WARNING**: Cloning the certrificate of the RDP server does not mean that the certificate will be trusted. Certificate trust requires a signed certificate from a CA that is **trusted** by the client. This is not likely to be the case in most scenarios. If you want to do that, you are on your own.
+> **WARNING**: Cloning the certificate of the RDP server does not mean that the certificate will be trusted. Certificate trust requires a signed certificate from a CA that is **trusted** by the client. This is not likely to be the case in most scenarios. If you want to do that, you are on your own.
 
 ## Steps
 
--- a/docs/cert-extraction.md
+++ b/docs/cert-extraction.md
@@ -1,0 +1,36 @@
+# Extracting Windows Server Remote Desktop Certificate
+
+This procedure is useful when running honeypots to support CredSSP (using `--auth ssp`).
+It requires Administrative privileges on the target server and the use of Mimikatz, so it assumes that you are able to deactivate the Anti-Virus on the target server.
+
+
+> **WARNING**: Cloning the certrificate of the RDP server does not mean that the certificate will be trusted. Certificate trust requires a signed certificate from a CA that is **trusted** by the client. This is not likely to be the case in most scenarios. If you want to do that, you are on your own.
+
+## Steps
+
+1. Turn off AV so mimikatz doesn't get flagged. (Or use excluded directory)
+2. Download mimikatz latest release.
+3. Go to `Start > Run... > certlm.msc` (optional)
+4. Identify the valid certificate under `Remote Desktop > Certificates` and note the thumbprint (optional)
+5. Export the Remote Desktop certificates using Mimikatz:
+
+   ```
+   crypto::capi
+   privilege::debug
+   crypto::cng
+   crypto::certificates /systemstore:LOCAL_MACHINE /store:"Remote Desktop" /export
+   ```
+
+6. Convert public key to `.pem` using openssl:
+
+   ```
+   openssl x509 -inform DER -outform PEM -in pubkey.der -out pubkey.pem
+   ```
+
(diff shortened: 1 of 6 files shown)


pyrdp/mitm/X224MITM.py
```diff
@@ -99,18 +99,24 @@ def onConnectionConfirm(self, pdu: X224ConnectionConfirmPDU):
...
+        # protocols = NegotiationProtocols.SSL if self.originalRequest.tlsSupported else NegotiationProtocols.NONE
 
         parser = NegotiationResponseParser()
         response = parser.parse(pdu.payload)
         if isinstance(response, NegotiationFailurePDU):
             self.log.info("The server failed the negotiation. Error: %(error)s", {"error": NegotiationFailureCode.getMessage(response.failureCode)})
             payload = pdu.payload
         else:
-            payload = parser.write(NegotiationResponsePDU(NegotiationType.TYPE_RDP_NEG_RSP, 0x00, protocols))
+            # payload = parser.write(NegotiationResponsePDU(NegotiationType.TYPE_RDP_NEG_RSP, 0x00, protocols))
```
alxbl, line 114: Oops, I think we still need to set the flags to `0x00` here otherwise the server might report that it supports Remote Admin, GFX, or a few other things that might cause the client to behave in an unintended way.

obilodeau:
> It also finally documents how to extract the private key of RDP servers so that we can close #154

That's awesome! :partying_face:

pyrdp/mitm/config.py
```diff
@@ -78,6 +79,8 @@ def __init__(self):
         self.useGdi: bool = False
         """Whether to allow the client to use the GDI rendering pipeline extension."""
 
+        self.authMethods: NegotiationProtocols = NegotiationProtocols.SSL
```
Res260, line 82: Dont forget do document this setting as well since all others have a docstring

pyrdp/mitm/cli.py
```diff
@@ -145,6 +146,7 @@ def buildArgParser():
                         default="INFO", choices=["INFO", "DEBUG", "WARNING", "ERROR", "CRITICAL"])
     parser.add_argument("-F", "--log-filter",
                         help="Only show logs from this logger name (accepts '*' wildcards)", default="")
+    parser.add_argument("--auth", help="Specify allowed authentication mechanisms (Comma-separated, choose from: tls, ssp)", default="tls")
```
Res260, line 149: Maybe not for this PR, but PyRDP supports the standard RDP security as well, so I think it should be included in the options.

docs/cert-extraction.md
```diff
@@ -0,0 +1,36 @@
+# Extracting Windows Server Remote Desktop Certificate
+
+This procedure is useful when running honeypots to support CredSSP (using `--auth ssp`).
+It requires Administrative privileges on the target server and the use of Mimikatz, so it assumes that you are able to deactivate the Anti-Virus on the target server.
+
+
+> **WARNING**: Cloning the certrificate of the RDP server does not mean that the certificate will be trusted. Certificate trust requires a signed certificate from a CA that is **trusted** by the client. This is not likely to be the case in most scenarios. If you want to do that, you are on your own.
```
Res260, line 7: typo certificate

pyrdp/mitm/cli.py
alxbl: None is always implicitly supported if the server selects NONE, hence why it's not on that list, and the default is TLS.

That being said, there's a bug (See X224MITM.py:119) where `startTLS` is done according to what the client requests, not what the server replies. I added a **FIXME** label because I didn't want to change the behavior of the code in this PR.

You also made me realize that in `onConnectionRequest` we should not blindly set the support auth methods, but rather, allowing only the configured ones through and setting everything else to false, otherwise if a non-CredSSP client connects, but `PyRDP` is started with `--auth ssp`, the server might mistakenly select SSP, which the client won't be able to respect.

PR #232, opened 2020-06-14
[WIP] Add basic support for parsing Dynamic Channels (drdynvc)

Res260:
While working on a CTF challenge, I wanted to test some things with the dynamic channels (if you come across this PR in the context of said CTF, this PR will NOT help you solve the challenge), so I built a basic MITM object to interact with them.

For reference, dynamic channels ( https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedyc/3bd53020-9b64-4c9a-97fc-90a79e7e1e06 ) is the "new and improved" way to extend RDP functions

Some examples of channels that were opened using win10 mstsc to another win10 machine:
image
image
image
image

I'll leave this PR as a WIP because it has not been tested enough to go to production and the current logs do not bring a lot of values. However, if we ever plan to implement a sub-MITM for a specific dynamic channel, this might come in handy!

### DIFF
--- a/pyrdp/layer/rdp/virtual_channel/dynamic_channel.py
+++ b/pyrdp/layer/rdp/virtual_channel/dynamic_channel.py
@@ -13,5 +13,5 @@ class DynamicChannelLayer(Layer):
     Layer to receive and send DynamicChannel channel (drdynvc) packets.
     """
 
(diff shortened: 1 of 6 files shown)


obilodeau:
Damn, I screwed up...

PR #241, opened 2020-07-27
Add RDPVersion 524301

sotebob:
builtins.ValueError: 524301 is not a valid RDPVersion
Add RDPVersion 524301


alxbl:
Cheers! Thanks for your contribution!

PR #248, opened 2020-10-02
fix: Clipboard file carving broke mapped drive transfer interception

### LINKED ISSUE #246: Clipboard Stealer / Mapped file transfer interception is broken
Looks like this was introduced by #100 

Looks like all of the file transfer interception is broken in master right now.

alxbl:
Closes #246


obilodeau (approved):
Looks good to me. Didn't test.

PR #249, opened 2020-10-02
fix: UTF-8 support for docker images.

### LINKED ISSUE #247: Crawler - Non-ASCII filenames cause exception to be thrown
Found out during live demo :) Callstack to come soon.

alxbl:
Closes #247.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # Handles compiling and package installation
 FROM ubuntu:18.04 AS compile-image
+
 # Install build dependencies
 RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 python3-pip \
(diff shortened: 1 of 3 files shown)


obilodeau (changes requested):
Changes in the Dockerfile seems to be only made to the compile image. Did you test this? If it fixed the problem it feels like the logger changes were enough.

That said, it might fail at the filesystem file open moment...

I think we should add the locale/utf-8 changes to the runtime image too.

alxbl:
Hmm, I just tested again, now I get utf-8... for both images (slim and not slim) I'm not sure why that is. I'll try to play around with the dockerfiles and see if we can just get rid of the extra steps.

obilodeau:
Have in mind that I bumped the docker images to ubuntu 20.04 recently in https://github.com/GoSecure/pyrdp/pull/251. Maybe there's a ubuntu fix or a python version change that fixed the behavior there.

alxbl:
Ah, that could be it. The `LANG` and `LC_TYPE` are still not set though.

In any case, I tried removing the Dockerfile changes and the issue still seems resolved. We'll keep the logging part in, but I believe that updating to 20.04 means Python 3.8, which uses utf8 by default.

obilodeau (approved):
LGTM, merging

Issue #250, opened 2020-10-06
pyrdp-convert.py fails in Docker container


robeving:
Run these commands
```
docker pull gosecure/pyrdp:latest
docker run -it gosecure/pyrdp /bin/bash
python3 /pyrdp/bin/pyrdp-convert.py
```

You get this error, missing dep in container?
```
Traceback (most recent call last):
  File "/pyrdp/bin/pyrdp-convert.py", line 15, in <module>
    from pyrdp.player.Mp4EventHandler import Mp4EventHandler
  File "/opt/venv/lib/python3.6/site-packages/pyrdp-0.4.2.dev0-py3.6-linux-x86_64.egg/pyrdp/player/Mp4EventHandler.py", line 14, in <module>
    import av
ModuleNotFoundError: No module named 'av'
```

Also:
`ls -l /pyrdp/bin/`

The pyrdp-convert script is missing the executable bit.

```
-rwxr-xr-x 1 root root  6776 Sep 14 18:25 pyrdp-clonecert.py
-rw-r--r-- 1 root root 15459 Sep 14 18:25 pyrdp-convert.py
-rwxr-xr-x 1 root root  1477 Sep 14 18:25 pyrdp-mitm.py
-rwxr-xr-x 1 root root  4004 Sep 14 18:25 pyrdp-player.py
```

obilodeau:
Fixed the missing executable bits in master minutes ago because it was really easy to do. 

Investigating the issue, I think I tracked it back to this PR #219 merging the Dockerfile improvements but we merged the pyrdp-convert changes after (although on a timeline it was done before I think). Our `requirements.txt` file is missing the dependencies.

I'll get to the bug shortly but we should exercise pyrdp-convert in CI as a smoke test too. @Res260 :eyes: if you have time to kill

obilodeau:
A candidate fix is available in #251.

obilodeau:
Successful regression testing added in #253

obilodeau:
A fix was merged minutes ago. Updated images are now available on DockerHub. Feel free to re-open if this doesn't fix your issue. Thanks for the report!

PR #251, opened 2020-10-07
pyrdp-convert.py fixes for #250

obilodeau:
* Bump to Ubuntu 20.04 in docker images
* Added back pyav deps to requirements.txt

Draft PR because I want to see how CI reacts.


obilodeau:
Added more checks in #253. Rebasing for a retest.

PR #254, opened 2020-10-12
Increase robustness of converting PCAPs with unhandled data

robeving:
When converting a *real* RDP session with pyrdp-convert it frequently fails to process PDUs with unhandled data. This is to be expected. If the exception is handled then conversion can generally continue without error producing a valid pyrdp output file.

### DIFF
--- a/bin/pyrdp-convert.py
+++ b/bin/pyrdp-convert.py
@@ -95,10 +95,13 @@ class RDPReplayer(RDPMITM):
         pass
 
     def recv(self, data: bytes, from_client: bool):
-        if from_client:
-            self.client.tcp.dataReceived(data)
-        else:
-            self.server.tcp.dataReceived(data)
+        try:
+            if from_client:
+                self.client.tcp.dataReceived(data)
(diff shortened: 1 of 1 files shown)


Res260 (commented):
Thanks for your PR @robeving ! I'll let @alxbl or @obilodeau confirm, but I think it'd be best if you used `self.log.error()` and/or `self.log.exception()` instead of `print()`, so the statement is treated the same as other log statements.

robeving:
> Thanks for your PR @robeving ! I'll let @alxbl or @obilodeau confirm, but I think it'd be best if you used `self.log.error()` and/or `self.log.exception()` instead of `print()`, so the statement is treated the same as other log statements.

As far as I can see print() is what is used in this file - although happy to change.

Res260:
You're right, I looked too quickly and saw the `self.log` property but it seems unused :p

alxbl (approved):
Hey @robeving, thanks for your contribution, this is very helpful!

PR #255, opened 2020-10-13
fix: bad logic for handling unsupported virtual channels

alxbl:
This is a PR following up #246 for the RDPDR delay that I noticed during regression testing.

The problem was flawed logic that would not create generic virtual channels when the channels were in the map but didn't have specific PyRDP parsers tied to them.


obilodeau (approved):
I gave this a good glance and some thoughts and it looks good.

PR #257, opened 2020-10-19
Feature: Add JSON conversion to `pyrdp-convert`

### LINKED ISSUE #236: Headless Player: Output event timestamps
Title is self-explanatory, the headless mode currently just dumps every event to `stdout` as though it had all happened at the same time, this is helpful for carving out malicious payloads or keystrokes, but discards timing info, which might be useful in some cases.

We should add support for outputting the timestamps around events.

An idea would be to group keystrokes within a certain time delta of each other into a single output line so as to not spam too much. Otherwise one timestamp per event would be fine.

----
Food for thought: we could even add some kind of `JSON` output format for `pyrdp-convert` eventually.

alxbl:
This refactors the conversion handler mechanism and adds a JSON conversion handler. I think it will be safer to land this after vNext, so the PR is only a draft right now. We can open and review it after the release.

TODO:
- [x] Test conversion on real captures.
- [x] Review JSON format to see if anything is missing
- [x] Release Notes

----
Closes #236.


obilodeau (commented):
* Quickly looked at the code, everything seemed fine.
* Quickly tested, it's working

I'm not sure about all these non-click mouse movement events. I wonder if they will ever be useful and not just spam the logs.

A question. With these events:
```json
    {
      "timestamp": 1583249189508,
      "type": "key",
      "data": {
        "key": "Shift",
        "press": true,
        "mods": []
      }
    },
    {
      "timestamp": 1583249189605,
      "type": "key",
      "data": {
        "key": "S",
        "press": true,
        "mods": []
      }
    },
    {
      "timestamp": 1583249189652,
      "type": "key",
      "data": {
        "key": "Shift",
        "press": false,
        "mods": []
      }
    },
```

Shouldn't the "S" have "shift" in mods? What's the logic for mods?

I might also have a file clipboard event that is interpreted as a text/plain clipboard event to show you. Can't share the capture publicly unfortunately.

alxbl:
Nice catch! You are right, the shift mod on S should be present. It looks like I forgot a bit of code to handle it.

For the mouse movement events, I figured they might be useful to detect automated input vs. human input, and I wanted to do an accurate conversion of the event stream (i.e. no artificial clean-up) since it's a format conversion. It shouldn't be too hard to filter them out post-facto.

alxbl:
So this is silly. I tab completed `scan` to `scancode` (module name) instead of `scanCode` (variable name). No key would ever be considered a modifier because of that. 🥇

obilodeau:
Ok for the mouse movements.

I'm glad I found a bug :smile: 

Feel free to merge this.

PR #258, opened 2020-10-19
Preparing 1.0.0 Release

alxbl:
I've followed the steps outlined in `docs/devel.adoc`.

I'd like a second set of eyes on the requirements.txt and setup.py versions, but I think they should be fine (nothing has changed since we added PyAV)

Once this is merged (squash merge) we can tag the release.


obilodeau (changes requested):
Good job!

Few things:
* I would plug that it is a SecTor release, like we did for previous releases where relevant
* I would add a paragraph saying it's 1.0 and we will avoid breaking backward compatibility in our CLI flags and featureset
* The CredSSP work is not in the list. I think we should go through all PRs to verify if another big thing was not forgotten. I try to always update the changelog right after merging PRs to avoid forgotting about important changes.

obilodeau (approved):
Discussed with Alex earlier today. He agreed if I tag the release.

obilodeau:
I rebased instead of squash since I wanted to keep the dependency updates separate.

Issue #261, opened 2020-11-13
Use hash as name for intercepted files


xshill:
It would be better if PyRDP used an intercepted file's hash instead of a random name when saving them to disk. Even though we already detect duplicate files using hashes and store the original file name in the `mapping.json` file, using hashes directly on the filesystem would make it easier for external tools to detect duplicates (e.g: if you're backing up intercepted files) because then they wouldn't have to parse the mapping file.

Issue #270, opened 2020-11-27
Standardize outputs of DeviceRedirectionMITM and FileCrawlerMITM


xshill:
Instead of being in `pyrdp_output/files`, they should be in `pyrdp_output/crawler` to make it easier to differentiate which files were intercepted and which files were crawled.

xshill:
On second thought, it would be much nicer to have a unified system for both components. Essentially:

- Both DeviceRedirectionMITM and FileCrawler should recreate the victim's filesystem in `pyrdp_output/filesystems/`
- Actual files should be written in `pyrdp_output/files/` and renamed to the sha1 hash of their contents
- Files in `filesystems/` should only link to the corresponding file in `files/`. On Windows, this could be a `.lnk` file or just a text file containing the path or the hash of the file (e.g: virus.exe would actually be called virus.exe.txt, it's content would be `[pyrdp_output_path]/files/sha1(virus.exe)`).
- All files should be logged in `pyrdp_output/mapping.json` (this implies changing the path attributes to arrays)

xshill:
One more note about `mapping.json`, we don't really need the `remote_path` attribute anymore since the filesystem is recreated. We can just have an array of `local_paths` were each file was saved.

xshill:
Actually I removed the `mapping.json` file since all the information from it can be gathered elsewhere and it makes the code much simpler.

Issue #268, opened 2020-11-19
Add tests for DeviceRedirectionMITM


xshill:
See #267.

Since the DeviceRedirectionMITM component needs changes anyway, this could be a good time for adding tests. Among the things that should be tested are:

- Duplicate files are deleted.
- Files are named after their hash (#261).
- All files get closed so we don't end up with too many open file descriptors.

xshill:
After experimenting a bit with `unittest.mock`, it seems much simpler to do than I originally thought.

Here are some things we would like to test:
- Make sure that client credentials are logged
- Make sure that responses to forged requests don't get sent to the server
- Make sure that open files are properly closed when a close request is received
- Make sure that read responses only handles files that we've opened
- Make sure that duplicate files are deleted
- Make sure that the file mapping contains proper info (e.g: sha1 hash, remote path)
- Make sure that the file mapping gets saved when it's supposed to

Issue #264, opened 2020-11-17
RDPDR File close responses are not handled properly by the MITM


xshill:
When transferring a file from a shared drive to the server filesystem, the RDPDR file stealer sometimes creates empty files (even though the file itself is not empty). It's useless and it clogs up the output folder and mapping.json file.

```
{
    "ChristineHerrera": {
        "creationTime": "2020-11-17T14:54:04.776912",
        "localPath": "pyrdp_output/files/ChristineHerrera",
        "remotePath": "\\Templates\\test.txt",
        "sha1": ""
    },
    "ClarenceMcanally": {
        "creationTime": "2020-11-17T14:55:47.534201",
        "localPath": "pyrdp_output/files/ClarenceMcanally",
        "remotePath": "\\Templates\\test.txt",
        "sha1": ""
    },
    "DavidRowles": {
        "creationTime": "2020-11-17T14:55:07.086586",
        "localPath": "pyrdp_output/files/DavidRowles",
        "remotePath": "\\Templates\\test.txt",
        "sha1": ""
    },
    "EvanDemaire": {
        "creationTime": "2020-11-17T15:03:44.626242",
        "localPath": "pyrdp_output/files/EvanDemaire",
        "remotePath": "\\Templates\\test.txt",
        "sha1": ""
    },
    "JesseWalker": {
        "creationTime": "2020-11-17T14:56:33.743618",
        "localPath": "pyrdp_output/files/JesseWalker",
        "remotePath": "\\Templates\\test.txt",
        "sha1": "c90e32d1e617ff4cb0ebd4789ded7ed10981a5e7"
    },
    "MaxKern": {
        "creationTime": "2020-11-17T14:55:18.115505",
        "localPath": "pyrdp_output/files/MaxKern",
        "remotePath": "\\Templates\\test.txt",
        "sha1": ""
    },
    "MichaelSummers": {
        "creationTime": "2020-11-17T14:55:57.601731",
        "localPath": "pyrdp_output/files/MichaelSummers",
        "remotePath": "\\Templates\\test.txt",
        "sha1": ""
    }
}
```

xshill:
The RDPDRMITM component doesn't handle file closes properly, which causes file duplications in file interceptor (among other problems). Although many file close responses are received, most of them don't make the check: `if key in self.openedFiles:`. Since they don't make it past that line, identical files never get filtered out and deleted by comparing hashes.

```
[...]

PR #272, opened 2020-12-23
File naming

xshill:
Closes issues #261, #264, #268, #270.

Most notably:

- Fix bug preventing files from being closed properly
- Standardize outputs of file crawler and device redirection component so that they both create a "clone" of the victim's filesystem
- Actual files are stored in the `files` folder and are named by hash for easy detection of duplicates
- The `filesystems` folder contains the filesystem clones, and files in this directory are symbolic links to the files in the `files` folder
- Symbolic links are created using the `Path.symlink_to` method in Python and the [documentation](https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to) suggests that this should work on both Linux and Windows
- Unit tests added for `DeviceRedirectionMITM` and `FileMapping`
- No more `mapping.json` file (the information it would contain should be easy to find anyway with the command line, and this simplifies the code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,7 @@ jobs:
         working-directory: ./
         run: coverage run --append -m unittest discover -v
 
-      - name: Coverage
+      - name: Coverage report
         working-directory: ./
         run: coverage report --fail-under=40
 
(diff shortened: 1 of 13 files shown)


obilodeau (approved):
Read the code. I'm not super familiar with what's going on here but nothing offended me. I think we should go with the merge. We will test before releasing so I think we will be good.

PR #273, opened 2021-01-04
Copy image before drawing cursor. Fixes #271.

### LINKED ISSUE #271: MP4 conversion - mouse pointer ghosts
Videos created by pyrdp-convert.py have mouse pointer ghosts effects. I think that we apply the next transformation on the image on which the pointer was painted. We should keep a copy of the original image and use that instead.

xshill:
When rendering an MP4 file, copy the image before drawing the cursor to avoid getting mouse pointer ghosts (see #271).

### DIFF
--- a/pyrdp/player/Mp4EventHandler.py
+++ b/pyrdp/player/Mp4EventHandler.py
@@ -132,8 +132,7 @@ class Mp4EventHandler(RenderingEventHandler):
     def _writeFrame(self, surface: QImage):
         w = self.stream.width
         h = self.stream.height
-        surface = surface.scaled(w, h) if self.scale else surface
-        frame = av.VideoFrame.from_image(ImageQt.fromqimage(surface))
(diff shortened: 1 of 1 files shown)


pyrdp/player/Mp4EventHandler.py
```diff
@@ -132,8 +132,7 @@ def onFinishRender(self):
     def _writeFrame(self, surface: QImage):
         w = self.stream.width
         h = self.stream.height
-        surface = surface.scaled(w, h) if self.scale else surface
-        frame = av.VideoFrame.from_image(ImageQt.fromqimage(surface))
```
obilodeau, line 136: The frame variable was not used?

xshill:
No, it was reassigned before being used

obilodeau (approved):
LGTM! You can merge. Also, the removal of the frame assignment might even make the conversions faster depending on how costly that operation was.

Issue #284, opened 2021-01-22
Improve data logging for parsing error


xshill:
When a parsing error occurs, we should log the data received at every layer instead of logging the raw TCP data. It would make it a lot easier to find the part that made the parsing fail.

PR #292, opened 2021-01-29
Write log showing the parser that crashed and the input that made it crash

xshill:
See issue #284


obilodeau (approved):
Looks good. Haven't tested a stacktrace. Did you?

PR #298, opened 2021-02-10
Use SO_REUSEADDR in pyrdp-mitm.py

### LINKED ISSUE #297: Use SO_REUSEADDR in pyrdp-mitm.py to avoid waiting for dead sockets
When sockets are stuck in a TIME_WAIT status, we have to wait some time for PyRDP to be able to bind to the same port again (long enough for it to be annoying). Using the SO_REUSEADDR socket option can fix the problem.

xshill:
Fix #297.


obilodeau (approved):
Feel free to merge :rocket: 

p.s.: don't forget changelog after

PR #299, opened 2021-03-03
Create parents when creating symlink's directory

xshill:
This fixes a bug causing the connection to crash when trying to save a file whose grand-parent directory hasn't been created yet.


obilodeau:
Is it safe to assume that the whole connection wasn't crashing but the individual IO channels were?

xshill:
> Is it safe to assume that the whole connection wasn't crashing but the individual IO channels were?

For me it made the whole connection crash on Windows.

PR #300, opened 2021-03-03
Log host ip

xshill:
Log value of HOST_IP environment variable when starting the MITM server. This will help to distinguish the log entries from various sources when they are all stashed together.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -23,6 +23,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 *MITM*
 
+* PyRDP will log the value of the HOST_IP variable when booting if it exists. You can set it to the IP address of the host running PyRDP. This is mostly helpful when you're using PyRDP in Docker and you want the IP of the Docker host in the logs.
 * Added detection function for BlueKeep scans / exploit attempts. PyRDP will log the attempt and shut down the connection. The JSON log has an exploitInfo attribute as well as a parserInfo attribute to help investigate what happened.
 * Added better logging for parser errors. PyRDP will now log which parser crashed and the data that was fed to that parser to make it crash. This makes it easier to investigate bugs and exploits. In JSON logs, this information shows up in the parserInfo attribute.
 * Files intercepted or crawled by the MITM are now named according to the sha1 hash of their contents and stored in the `pyrdp_output/files` folder (see {uri-issue}261[#261])
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -23,7 +23,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 *MITM*
 
-* PyRDP will log the value of the HOST_IP variable when booting if it exists. You can set it to the IP address of the host running PyRDP. This is mostly helpful when you're using PyRDP in Docker and you want the IP of the Docker host in the logs.
+* PyRDP will log the value of the `HOST_IP` variable on start if it exists. You can set it to the IP address of the host running PyRDP. This is mostly helpful when you're using PyRDP in Docker and you want the IP of the Docker host in the logs.
 * Added detection function for BlueKeep scans / exploit attempts. PyRDP will log the attempt and shut down the connection. The JSON log has an exploitInfo attribute as well as a parserInfo attribute to help investigate what happened.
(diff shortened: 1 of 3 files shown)


Res260:
Isn't this redundant with the --sensor-id argument? Sorry if I missed something

xshill:
> Isn't this redundant with the --sensor-id argument? Sorry if I missed something

We actually don't use them the same way in our internal stuff.

bin/pyrdp-mitm.py
```diff
@@ -43,6 +44,9 @@ def main():
 
     logger.info("MITM Server listening on port %(port)d", {"port": config.listenPort})
 
+    if "HOST_IP" in os.environ:
+        logger.info("Host IP: %(host_ip)s", {"host_ip": os.environ["HOST_IP"]})
```
obilodeau, line 48: I think both messages should be in the same log. This will greatly simplify event processing later for a little cost.

README.md
```diff
@@ -563,6 +563,14 @@ docker run -v $PWD/pyrdp_output:/home/pyrdp/pyrdp_output -p 3389:3389 gosecure/p
 Make sure that your destination directory is owned by a user with a UID of 1000, otherwise you will get permission denied errors.
 If you are the only non-root user on the system, usually your user will be assigned UID 1000.
 
+#### Logging the host IP address
+
+If you want PyRDP to log the host IP address in its logs, you can set the HOST_IP environment variable when using `docker run`:
+
+```
+docker run -p 3389:3389 -e HOST_IP=192.168.1.9 gosecure/pyrdp pyrdp-mitm.py 192.168.1.10
```
obilodeau, line 571: I'm not sure about the HOST_IP name. Does this align to something you've seen used elsewhere with containers?

obilodeau (changes requested):
Straightforward and well done but some changes needed.

README.md
xshill: No, we could name it anything else honestly.

bin/pyrdp-mitm.py
xshill: sure

README.md
obilodeau: After thinking about it a little more, `HOST_IP` is fine.

CHANGELOG.adoc
```diff
@@ -23,6 +23,7 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/master[
 
 *MITM*
 
+* PyRDP will log the value of the HOST_IP variable when booting if it exists. You can set it to the IP address of the host running PyRDP. This is mostly helpful when you're using PyRDP in Docker and you want the IP of the Docker host in the logs.
```
obilodeau, line 26:
```suggestion
* PyRDP will log the value of the `HOST_IP` variable on start if it exists. You can set it to the IP address of the host running PyRDP. This is mostly helpful when you're using PyRDP in Docker and you want the IP of the Docker host in the logs.
```

README.md
```diff
@@ -563,6 +563,14 @@ docker run -v $PWD/pyrdp_output:/home/pyrdp/pyrdp_output -p 3389:3389 gosecure/p
 Make sure that your destination directory is owned by a user with a UID of 1000, otherwise you will get permission denied errors.
 If you are the only non-root user on the system, usually your user will be assigned UID 1000.
 
+#### Logging the host IP address
+
+If you want PyRDP to log the host IP address in its logs, you can set the HOST_IP environment variable when using `docker run`:
```
obilodeau, line 568:
```suggestion
If you want PyRDP to log the host IP address in its logs, you can set the `HOST_IP` environment variable when using `docker run`:
```

bin/pyrdp-mitm.py
```diff
@@ -41,7 +43,14 @@ def main():
     reactor.adoptStreamPort(s.fileno(), socket.AF_INET, MITMServerFactory(config))
     s.close()  # reactor creates a copy of the fd.
 
-    logger.info("MITM Server listening on port %(port)d", {"port": config.listenPort})
+    message = "MITM Server listening on port %(port)d"
```
obilodeau, line 46: Stating the bind-address in the logs is more accurate and might prevent surprises from users since HOST_IP != how we bind:

```suggestion
    message = "MITM Server listening on 0.0.0.0:%(port)d"
```

obilodeau (commented):
A couple of quick suggestions.

obilodeau (approved):
:+1: :rocket: :musical_keyboard:

PR #301, opened 2021-03-12
Update README.md

exys228:
Fixed quotes in Windows readme section, gave me a little bit of a headache


xshill:
Just curious, was the use of single quotes causing a specific technical problem on Windows cmd?

xshill:
Ok I see the problem, thanks for the fix :smile:

PR #303, opened 2021-03-14
Log cred attempts when NLA is not used

xshill:
- Handle space in password attempts
- Log attempt when enter is pressed, not just when we receive the login PDU
- Log attempt when the mouse is clicked in a predefined portion of the screen (currently between 50% and 65% of width and height)


pyrdp/mitm/FastPathMITM.py
```diff
@@ -3,11 +3,12 @@
 # Copyright (C) 2019 GoSecure Inc.
```
obilodeau, line 3: copyright bump

obilodeau (approved):
This is great!

PR #305, opened 2021-03-18
Copyright bump to 2021

obilodeau:
Some files were modified further. I went through all changed files with `git diff HEAD HEAD@{2020-12-31} --name-only` and verified individual history with `git log -p <filename>`.

### DIFF
--- a/pyrdp/layer/layer.py
+++ b/pyrdp/layer/layer.py
@@ -1,6 +1,6 @@
 #
 # This file is part of the PyRDP project.
-# Copyright (C) 2018 GoSecure Inc.
+# Copyright (C) 2018, 2019, 2021 GoSecure Inc.
 # Licensed under the GPLv3 or later.
 #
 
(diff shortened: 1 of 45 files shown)


pyrdp/layer/layer.py
```diff
@@ -1,6 +1,6 @@
 #
 # This file is part of the PyRDP project.
-# Copyright (C) 2018 GoSecure Inc.
+# Copyright (C) 2018, 2019, 2021 GoSecure Inc.
```
xshill, line 3: Can't we just write 2018-2021? Here and a couple other places.

xshill: (basically everywhere we have more than 1 year)

xshill (changes requested):

pyrdp/layer/layer.py
obilodeau: Unfortunately no, where I wrote without dashes are places where there were no code changes in the missing years. I'm not sure if we can apply the copyright statement if there was no change.

xshill: ah ok then

PR #307, opened 2021-03-24
Log NTLMSSP hashes when using NLA

xshill:
Log NTLMSSP hashes in a format usable by John the Ripper and Hashcat when the client uses NLA.

NTLMSSP hashes are stored in `pyrdp_output/logs/ntlmssp.log` and are also logged to the console and in JSON format.

This might actually be usable for gathering hashes in pentest scenarios, but I haven't tried anything like that yet.


obilodeau:
LGTM! Great work! Can't wait to see it in action 🙂

obilodeau (approved):

Issue #260, opened 2020-11-04
Add option to detect NLA and optionally redirect to a non NLA server.


alxbl:
When the server enforces NLA (in NEG_RSP) it should be possible to tell PyRDP to ignore the real server and connect to a malicious server for credential collection at the very least.

PR #308, opened 2021-03-25
NLA redirection

xshill:
Added options `--redirection-host` and `--redirection-port`. When trying to downgrade from NLA, PyRDP will connect to this host instead of the intended server if it only allows NLA connections.

See #260.

### DIFF
--- a/pyrdp/mitm/RDPMITM.py
+++ b/pyrdp/mitm/RDPMITM.py
@@ -206,7 +206,7 @@ class RDPMITM:
 
         await serverFactory.connected.wait()
 
-        if self.config.attackerHost is not None and self.config.attackerPort is not None:
+        if self.config.attackerHost is not None and self.config.attackerPort is not None and not self.player.tcp.connectedEvent.is_set():
             attackerFactory = AwaitableClientFactory(self.player.tcp)
             reactor.connectTCP(self.config.attackerHost, self.config.attackerPort, attackerFactory)
 
--- a/pyrdp/mitm/RDPMITM.py
+++ b/pyrdp/mitm/RDPMITM.py
@@ -193,9 +193,9 @@ class RDPMITM:
 
         serverFactory = AwaitableClientFactory(self.server.tcp)
         if self.config.transparent:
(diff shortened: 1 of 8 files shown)


pyrdp/mitm/RDPMITM.py
```diff
@@ -177,6 +181,10 @@ def getServerLog(self, name: str) -> SessionLogger:
         """
         return self.serverLog.createChild(name)
 
+    def disconnectFromServer(self):
+        self.server.replaceTCP()
+        self.tcp.setServer(self.server.tcp)
```
obilodeau, line 186: why is this required?

pyrdp/mitm/cli.py
```diff
@@ -127,6 +127,8 @@ def buildArgParser():
         "--transparent", help="Spoof source IP for connections to the server (See README)", action="store_true")
     parser.add_argument("--no-gdi", help="Disable accelerated graphics pipeline (MS-RDPEGDI) extension",
                         action="store_true")
+    parser.add_argument("--redirection-host", help="Redirection target ip if NLA is enforced", default=None)
+    parser.add_argument("--redirection-port", help="Redirection target port if NLA is enforced", type=int, default=None)
```
obilodeau, line 131: I think we should name the flags `--nla-redirection-host` and `--nla-redirection-port` to make explicit that it is used when NLA nego fails.

obilodeau (commented):
I have concerns about the interaction of this code with the transparent proxy mode. Are both modes compatible? It should be possible but how will the new server-side TCP connection will be created? It needs to be set with clientIP as its source. Is that covered?

xshill:
> I have concerns about the interaction of this code with the transparent proxy mode. Are both modes compatible? It should be possible but how will the new server-side TCP connection will be created? It needs to be set with clientIP as its source. Is that covered?

I suppose so, it uses the same functions as previously, just with a different argument. But I haven't tested it tbh.

obilodeau (approved):
Looks good to me :rocket:

Issue #265, opened 2020-11-18
Regression: 'Too Many Files Open' in MITM component


alxbl:
It looks like #48 has croppped up again in the MITM component. We should spend some time figuring out how to trace open handles.

A good way to reproduce quickly seems to be to do a SYN flood on the PyRDP instance. (Maybe try masscan or write a simple asyncio flooder in Python)

We'd need to investigate how to debug file descriptor leaks and determine if this is a leak or just denial of service.

Off the top of my head, the source of leak/exhaustion might be:

- Session Log files
- Socket Descriptors (inbound, certificate request, outbound)
- Replay file sink

One successful RDP connection would generate:
- Session Log (2x)
- JSON Session Log (2x)
- Replay File (2x)
- Mapping.json file (1x)
- 4 socket descriptors (in/out x2)

The reason these numbers are doubled is because of the initial certificate request which causes a disconnect

**Things to check:**

- Does Twisted clean up connections?
- Do we properly close log handles upon client disconnection?
- Do we have reasonable connection timeouts for dangling SYNs?
- Do we have a reasonable accept queue for twisted?

obilodeau:
I think @xshill identified the issue. See #264.

alxbl:
There's a good chance that the RDPDR stuff is part of the issue, but it looks like it's possible to reproduce this behaviour without any file transfers or mapped drives, so I doubt that's all there is to it.

obilodeau:
I'll repro the "too many files open" stack trace with a scanner. Then test if @xshill's fix takes care of that for simple cases at least. We should plug the leaks but I don't think we should lose too much time trying to survive 10,000 concurrent connections.

obilodeau:
Tested with `rdpassspray` and `masscan` this morning and couldn't crash pyrdp. However, these tools are gentle: rdpassspray spawns a xfreerdp and wait on it to finish before retrying and masscan wait 1 second on a retry on the same host/port combo.

Googling for nmap and stress testing I stumbled upon: https://www.rapid7.com/db/modules/auxiliary/dos/tcp/synflood/. Using this doesn't negatively affects PyRDP on current master branch. The SYN connections don't even trigger logging.

For file transfers, @yraiesets did whole directory transfers stress tests with and without executables in them and nothing wrong happened.

I think the file mapping changes and the linked issues fixed this. I'm closing this. Let's open a new ticket if we have a clear reproducer.

PR #309, opened 2021-03-31
Watch for disconnection events in DeviceRedirectionMITM

xshill:
Watch for disconnection events and make sure to close files when either the client or server disconnects.

This should help with #265.

### DIFF
--- a/pyrdp/mitm/FileMapping.py
+++ b/pyrdp/mitm/FileMapping.py
@@ -33,7 +33,8 @@ class FileMapping:
         self.written = False
 
     def seek(self, offset: int):
-        self.file.seek(offset)
+        if not self.file.closed:
+            self.file.seek(offset)
 
     def write(self, data: bytes):
         self.file.write(data)
@@ -54,6 +55,9 @@ class FileMapping:
         return sha1.hexdigest()
 
     def finalize(self):
+        if self.file.closed:
+            return
+
         self.log.debug("Closing file %(path)s", {"path": self.dataPath})
         self.file.close()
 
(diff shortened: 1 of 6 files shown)


pyrdp/mitm/FileMapping.py
```diff
@@ -82,6 +86,11 @@ def finalize(self):
                 "path": str(self.filesystemPath.relative_to(self.filesystemDir)), "shasum": fileHash
             })
 
+    def onDisconnection(self, reason):
+        if not self.file.closed:
+            self.file.close()
+            Path(self.file.name).unlink(missing_ok=True)
```
obilodeau, line 92: why are we removing files on disconnection? Is it because they would be incomplete?

test/test_X224MITM.py
```diff
@@ -11,7 +11,7 @@
 from pyrdp.pdu import X224ConnectionRequestPDU, NegotiationRequestPDU
 
 
-class FileMappingTest(unittest.TestCase):
+class X224MITMTest(unittest.TestCase):
```
obilodeau, line 14: :foot: :beetle: :heavy_check_mark:

obilodeau (commented):
Looks good. Just one question.

pyrdp/mitm/FileMapping.py
xshill: Yes, if they were complete the `finalize` function would have been called and `self.file.closed` would be `True`.

obilodeau: Wouldn't we want to keep partial transfers? Maybe label them as such. Let's say we are monitoring a session where someone uploads a large zip or a VMDK. We could still extract intel from the file even if it's incomplete. Just a thought.

Issue #352, opened 2021-08-30
PyRDP Convert to MP4: out of memory issues and corrupted video


obilodeau:
I can reliably make the `pyrdp-convert.py` tool go out of memory:

```
$ time pyrdp-convert.py rdp_replay_20210826_12-18-46_415_Kimberly835337.pyrdp -f mp4
[*] Converting 'rdp_replay_20210826_12-18-46_415_Kimberly835337.pyrdp' to MP4
 45% (241939 of 536390) |######################                           | Elapsed Time: 8:37:30 ETA:   6:00:53Terminated

real	517m44.110s
user	765m28.181s
sys	10m30.343s
```

My OOM logs say:
```
Aug 27 21:01:27 barachois earlyoom[1099]: mem avail:  3180 of 31959 MiB ( 9.95%), swap free: 6125 of 6143 MiB (99.69%)
Aug 27 21:01:27 barachois earlyoom[1099]: low memory! at or below SIGTERM limits: mem 10.00%, swap 100.00%
Aug 27 21:01:27 barachois earlyoom[1099]: sending SIGTERM to process 526823 uid 1000 "pyrdp-convert.p": badness 848, VmRSS 10362 MiB
Aug 27 21:01:27 barachois earlyoom[1099]: process exited after 0.1 seconds
```

So 10GB of RAM for a 1GB replay file:
```
$ ls -lh rdp_replay_20210826_12-18-46_415_Kimberly835337.pyrdp 
-rw-r--r-- 1 olivier olivier 973M Aug 26 21:18 rdp_replay_20210826_12-18-46_415_Kimberly835337.pyrdp
```

@alxbl created patches to mitigate these issues. I'll take a look at them.

Additionally, the converted MP4 is unplayable, its `moov` atom was not written out to disk so the resulting file is useless. I will try to address that problem separately by using a streaming container format instead of a fixed MP4.

obilodeau:
https://ffmpeg.org/ffmpeg-formats.html#mov_002c-mp4_002c-ismv:

[...]

PR #353, opened 2021-08-30
Reduce memory usage of pyrdp-convert.py's conversion to video

obilodeau:
The first change is to use an MP4 streaming format instead of the default MP4 encoder.

The second makes our BitmapCache track the QImage backing buffer in addition to the QImage itself.

The last one plugs a memory leak in the run length encoding decoding routine in our C extension `rle.c`.

See #352 for the whole story including profiling and performance testing.


obilodeau:
Also see #352.

Tests are in-progress. Will post results once done.

PR #362, opened 2021-09-18
PyRdp-Player optimize for Mac OSX

thelongestusernameofall:
Bugfix: 1. fix Memory Not Enough Error (caused by Circular import) by line 37 modification;
    2. fix PySide2 freeze Error by line 28 modification;
    After these modification, pyrdp-player works well on mac osx.

### DIFF
--- a/bin/pyrdp-player.py
+++ b/bin/pyrdp-player.py
@@ -25,6 +25,7 @@ import logging  # noqa
 import logging.handlers  # noqa
 import sys  # noqa
 import os  # noqa
+# Workaround a macOS bug: https://bugreports.qt.io/browse/QTBUG-87014
 os.environ['QT_MAC_WANTS_LAYER']='1'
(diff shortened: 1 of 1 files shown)


obilodeau:
Great contribution! Thanks!

bin/pyrdp-player.py
```diff
@@ -25,7 +25,7 @@
 import logging.handlers  # noqa
 import sys  # noqa
 import os  # noqa
-
+os.environ['QT_MAC_WANTS_LAYER']='1'
```
obilodeau, line 28:
```suggestion
# Workaround a macOS bug: https://bugreports.qt.io/browse/QTBUG-87014
os.environ['QT_MAC_WANTS_LAYER']='1'
```

obilodeau (commented):
Comment suggestion

PR #364, opened 2021-10-24
fix(359): Support fractional fps.

### LINKED ISSUE #359: Increase/validate precision of converted videos
I believe that the way we encode frames we create the possibility of a drift between the real time and the resulting video times. This is exacerbated on large captures.

In Mp4EventHandler's init:
```python
        self.delta = 1000 // fps  # ms per frame
```
then:
```python
        nframes = (dt // self.delta)
        if nframes > 0:
            for _ in range(nframes):
                self.writeFrame()
```

However:
```ipython
In [18]: 1000//30
Out[18]: 33

In [19]: 1000/30
Out[19]: 33.333333333333336
```

There's a 0.3ms per 1000ms drift opportunity. On a long capture this amount to a very large maximum theoretical drift:
```ipython
In [26]: ((1000/30-1000//30) * 12*60*60) / 60
Out[26]: 240.0000000000017
```

Further investigation: I will record an RDP session playing a high precision clock and timer on a screen for a long time and compare after video encoding.

alxbl:
This pull request fixes #359

I've done two things:

- Default FPS is now 25. (PS: It looks like a recent refactor makes it impossible to configure the FPS for the MP4 converter)
- Use fractional division and keep track of the drift. Whenever we drift by more than one frame, an extra frame is added.

I'm not sure if the drift was a real issue though since we only updated the previous timestamp whenever at least one frame was rendered. In other words, if the timestamp between 2 PDUs is smaller than the frame rate delta, we would just let the delta increase until we've gone past the next integer frame, then render the screen at the current PDU.

The only thing that might happen is that at low frame rates, some information is lost because the screen changes faster than our frame rate can observe. RDP doesn't really have a concept of framerate, updates are sent as they happen, so it's difficult to map it to a constant frame rate.

### DIFF
--- a/pyrdp/convert/MP4EventHandler.py
+++ b/pyrdp/convert/MP4EventHandler.py
@@ -89,13 +89,14 @@ class MP4EventHandler(RenderingEventHandler):
 
         # Prevent drifting when fps doesn't perfectly divide a second.
         nframes = (dt / self.delta) + self.drift
-        self.drift = nframes - floor(nframes)  # update drift.
+        nwhole = floor(nframes)
+        self.drift = nframes - nwhole  # update drift.
 
-        if floor(nframes) > 0:  # Only take whole frames.
-            for _ in range(nframes):
+        if nwhole > 0:  # Only take whole frames.
+            for _ in range(nwhole):
                 self.writeFrame()
             self.prevTimestamp = ts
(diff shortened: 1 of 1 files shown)


obilodeau (changes requested):
Marking this clearly has "needing changes" to avoid accidental merge.

PR #365, opened 2021-10-25
chore(363): replace notify2 by py-notifier

### LINKED ISSUE #363: notify2 library is deprecated
Read about it here: https://bitbucket.org/takluyver/pynotify2/src/master/

alxbl:
This replaces notify2 by py-notifier.

I think (but would need further testing) that this allows us to get rid of dbus dependencies since it uses `libnotify-bin`. Which has been added to the README and docker files (not relevant to the CI pipeline AFAIK)

I will investigate whether dbus can be removed and probably open a second PR to cleanup the dependencies.

This also paves the road to eventually supporting windows notifications once win10toast is fixed. (see https://github.com/jithurjacob/Windows-10-Toast-Notifications/issues/86)

Closes #363.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,6 +48,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends python3 \
         # GUI and notifications stuff
         libgl1-mesa-glx libxcb-xinerama0 \
         notify-osd dbus-x11 libxkbcommon-x11-0 \
+        libnotify-bin \
         # Runtime requirement for PyAV (pyrdp-convert to MP4)
         libavcodec58 libavdevice58 \
         && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -9,9 +9,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         python3-setuptools \
         # Required for venv setup
         python3-venv \
-        # Required to build RLE module and dbus-python (GUI)
+        # Required to build RLE module (GUI)
```
obilodeau, line 12:
```suggestion
        # Required to build RLE module
```
The `GUI` mention was referring to dbus-python (a GUI related dependency). Needs to be removed now.

obilodeau (commented):
One last little thing. See comment below. Also, I'm off to test in a Ubuntu server system now w/o a GUI and see if my assumptions about `notify-osd` are correct.

Issue #358, opened 2021-09-03
Grab the Net-NTLM hash before closing an NLA connection


obilodeau:
We know that [Responder](https://github.com/lgandx/Responder) is [able to dump Net-NTLMv2 hashes](https://twitter.com/pythonresponder/status/1162814428850589697) even if it doesn't fully support RDP.

This means that if we pursue the handshake even if we know we can't finish the connection, we can at least grab the hash and log it.

A stretch goal would be: if we manage to crack the hash, we know we could MITM NLA because we could then create a new challenge-response on the server side. If I recall correctly the only thing preventing us from doing that was that part of the challenge response mixed the plaintext password (which we don't know) with the server's public/private/fingerprint (not sure which) and that the server would reject anything tampered. We couldn't do the double diffie-hellman trick because of the mixing of both these layers. If we have the password, we can truly do an NLA handshake in the middle. This would open up a new attack use case. Note that I'm half intentionally vague here.

Tagging @danielsantos1234.

PR #367, opened 2021-10-26
(#358) Catch NTLM hash if non-SSP authentication

lubiedo:
This PR will enable PyRDP to catch the NTLM Hash during a client's connection even when it was not executed with CredSSP authentication enabled and the server enforces it. To achieve this it will cut connection with the server and create a challenge PDU for the client to respond with an authentication message.

cc @obilodeau


lubiedo:
Getting closer to finishing this PR. @obilodeau: let me know what you think and if you have some changes for me to add to this.

pyrdp/security/nla.py
```diff
@@ -22,7 +24,7 @@ class NLAHandler(SegmentationObserver):
     This also logs the hash of NLA connection attempts.
     """
 
-    def __init__(self, sink: IntermediateLayer, state: NTLMSSPState, log: logging.LoggerAdapter):
+    def __init__(self, sink: IntermediateLayer, state: NTLMSSPState, log: logging.LoggerAdapter, ntlmCatch: bool = False):
```
obilodeau, line 27: `ntlmCatch` here should probably be turned into a keyword argument. For example, if we want to add NTLM relay capability here, the current api would look like:

```
NLAHandler(sink, state, log, False, True)
```
The above being `ntlmCatch=False`, `ntlmRelay=True`.

Whereas if we use keyword arguments it could be:
```
NLAHandler(sink, state, log, relay=True)
```

Maybe using keyword arguments is too complex for a simple use case like that. I wonder what pattern already exists in PyRDP that could allow this Object to be extended cleanly. We need to check.

Also lets change `ntlmCatch` for `ntlmCapture` or `captureNtlm` idk.

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -86,3 +92,109 @@ def parseNTLMSSPAuthenticate(self, data: bytes, stream: BytesIO) -> NTLMSSPAuthe
...
+        substream.write(Uint8.pack(NTLMSSPChallengeVersion.NEG_PROD_MAJOR_VERSION_HIGH))   # product major version
+        substream.write(Uint8.pack(NTLMSSPChallengeVersion.NEG_PROD_MINOR_VERSION_LOW))    # product minor version
+        substream.write(Uint16LE.pack(NTLMSSPChallengeVersion.NEG_PROD_VERSION_BUILT))     # product build
+        substream.write(Uint8.pack(0))                                                     # reserved
+        substream.write(Uint8.pack(0))                                                     # reserved
+        substream.write(Uint8.pack(0))                                                     # reserved
+        substream.write(Uint8.pack(NTLMSSPChallengeVersion.NEG_NTLM_REVISION_CURRENT))     # NTLM revision current
+        
+        self.writeNTLMSSPTSRequest(stream, NTLMSSPChallengeVersion.CREDSSP_VERSION, substream.getvalue())
+        return stream.getvalue()
```
obilodeau, line 156: I'm going to close this review because there is already a couple of items to fix and look at but I wonder if we can simplify this pattern somehow. We should check for other code examples where long sequences of bytes are written like that.

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -86,3 +92,109 @@ def parseNTLMSSPAuthenticate(self, data: bytes, stream: BytesIO) -> NTLMSSPAuthe
...
+
+    def writeNTLMSSPTSRequest(self, stream: BytesIO, version: int, negoTokens: bytes):
```
obilodeau, line 158: add link to MS documentation here too

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -86,3 +92,109 @@ def parseNTLMSSPAuthenticate(self, data: bytes, stream: BytesIO) -> NTLMSSPAuthe
...
+        stream.write(ber.writeInteger(version)) # CredSSP version
+        stream.write(ber.writeContextualTag(1, negoLen + 16))
+        stream.write(ber.writeUniversalTag(ber.Tag.BER_TAG_SEQUENCE, True))
+        stream.write(ber.writeLength(negoLen + 12))
+        stream.write(ber.writeUniversalTag(ber.Tag.BER_TAG_SEQUENCE, True))
+        stream.write(ber.writeLength(negoLen + 8))
+        stream.write(ber.writeContextualTag(0, negoLen + 4))
+        stream.write(ber.writeOctetString(negoTokens))
+
+    def writeNTLMSSPChallengePayload(self, stream: BytesIO, workstation: str) -> int:
```
obilodeau, line 173: add link to documentation

pyrdp/pdu/rdp/ntlmssp.py
```diff
@@ -13,17 +17,19 @@ def __init__(self, messageType: NTLMSSPMessageType):
         super().__init__()
         self.messageType = messageType
 
-
```
obilodeau, line 16: PEP8 says use two newline to separate classes, this change should be reverted

pyrdp/pdu/rdp/ntlmssp.py
```diff
@@ -13,17 +17,19 @@ def __init__(self, messageType: NTLMSSPMessageType):
...
 class NTLMSSPNegotiatePDU(NTLMSSPPDU):
     def __init__(self):
         super().__init__(NTLMSSPMessageType.NEGOTIATE_MESSAGE)
 
-
```
obilodeau, line 21: same as above

pyrdp/pdu/rdp/ntlmssp.py
```diff
@@ -13,17 +17,19 @@ def __init__(self, messageType: NTLMSSPMessageType):
...
 class NTLMSSPChallengePDU(NTLMSSPPDU):
     def __init__(self, serverChallenge: bytes):
         super().__init__(NTLMSSPMessageType.CHALLENGE_MESSAGE)
         self.serverChallenge = serverChallenge
 
+class NTLMSSPChallengePayloadPDU(PDU):
+    def __init__(self, workstation: str):
+        super().__init__()
+        self.workstation = workstation
```
obilodeau, line 33: missing newline for PEP8

pyrdp/pdu/rdp/ntlmssp.py
```diff
@@ -32,3 +38,13 @@ def __init__(self, user: str, domain: str, proof: bytes, response: bytes):
         self.domain = domain
         self.proof = proof
         self.response = response
+
```
obilodeau, line 41: missing newline for PEP8

pyrdp/pdu/rdp/ntlmssp.py
```diff
@@ -13,17 +17,19 @@ def __init__(self, messageType: NTLMSSPMessageType):
...
     def __init__(self):
         super().__init__(NTLMSSPMessageType.NEGOTIATE_MESSAGE)
 
-
 class NTLMSSPChallengePDU(NTLMSSPPDU):
     def __init__(self, serverChallenge: bytes):
         super().__init__(NTLMSSPMessageType.CHALLENGE_MESSAGE)
         self.serverChallenge = serverChallenge
 
+class NTLMSSPChallengePayloadPDU(PDU):
```
obilodeau, line 29: link to MS documentation

pyrdp/pdu/rdp/ntlmssp.py
```diff
@@ -32,3 +38,13 @@ def __init__(self, user: str, domain: str, proof: bytes, response: bytes):
         self.domain = domain
         self.proof = proof
         self.response = response
+
+class NTLMSSPTSRequestPDU(PDU):
+    """
+    PDU for TSRequest structures used by CredSSP (client/server) for SPNEGO and Kerberos/NTLM messages
```
obilodeau, line 44: link to MS documentation

pyrdp/security/nla.py
```diff
@@ -33,15 +35,35 @@ def __init__(self, sink: IntermediateLayer, state: NTLMSSPState, log: logging.Lo
         self.sink = sink
         self.ntlmSSPState = state
         self.log = log
+        self.catch = ntlmCatch
```
obilodeau, line 38: catch -> capture

pyrdp/security/nla.py
```diff
@@ -33,15 +35,35 @@ def __init__(self, sink: IntermediateLayer, state: NTLMSSPState, log: logging.Lo
...
         self.ntlmSSPParser = NTLMSSPParser()
 
+    def getRandChallenge(self):
```
obilodeau, line 41: Should we use the server-provided random values? I'm afraid we are going to introduce a vulnerability here by using a random number generator that is known to be weak: 
image

from https://docs.python.org/3/library/random.html

obilodeau (commented):
Things are taking shape. Here's my review. A lot of little things, nothing major. Great work!

pyrdp/core/ber.py
```diff
@@ -140,7 +140,27 @@ def writeApplicationTag(tag: Tag, size: int) -> bytes:
         return Uint8.pack((Class.BER_CLASS_APPL | PC.BER_CONSTRUCT) | Tag.BER_TAG_MASK) + Uint8.pack(tag) + writeLength(size)
     else:
         return Uint8.pack((Class.BER_CLASS_APPL | PC.BER_CONSTRUCT) | (Tag.BER_TAG_MASK & tag)) + writeLength(size)
-    
+
+def readContextualTag(s: BinaryIO, tag: Tag, isConstruct: bool) -> bool:
+    """
+    Unpack contextual tag and return True if the proper tag was read.
```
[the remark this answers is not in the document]
lubiedo: You are correct. Fixing this.

pyrdp/mitm/X224MITM.py
```diff
@@ -117,19 +117,27 @@ def onConnectionConfirm(self, pdu: X224ConnectionConfirmPDU):
...
+                    self.state.useRedirectionHost()
+                    self.onConnectionRequest(self.originalConnectionRequest)
+                    return
+                else:
+                    # Respond with a RDP negotiation answer to proceed and catch NTLM
+                    self.log.info("Server requires CredSSP. Closing connection with server and attempting to catch NTML hashes.")
+                    payload = parser.write(NegotiationResponsePDU(NegotiationType.TYPE_RDP_NEG_RSP, 0x00, NegotiationProtocols.CRED_SSP))
+                    self.state.ntlmCatch = True
+
+                    self.disconnector()
```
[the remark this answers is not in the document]
lubiedo: In this case we don't want to return because, unlike the clause above, we are not redirecting the connection but continuing with the SSL connection established on line 149.

pyrdp/security/nla.py
lubiedo: Please, check commit 20fa506 and let me know what you think.

pyrdp/mitm/RDPMITM.py
```diff
@@ -238,8 +239,11 @@ def doClientTls(self):
 
         # Add unknown packet handlers.
         ntlmSSPState = NTLMSSPState()
-        self.client.segmentation.addObserver(NLAHandler(self.server.tcp, ntlmSSPState, self.getLog("ntlmssp")))
-        self.server.segmentation.addObserver(NLAHandler(self.client.tcp, ntlmSSPState, self.getLog("ntlmssp")))
+        if self.state.ntlmCatch:
+            self.client.segmentation.addObserver(NLAHandler(self.client.tcp, ntlmSSPState, self.getLog("ntlmssp"), True))
```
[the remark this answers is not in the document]
lubiedo: Please, check commit 20fa506 and let me know what you think.

pyrdp/security/nla.py
lubiedo: The RDP server is the one in charge of creating the challenge value so yes, it should be created by PyRDP in this case. Either way, good catch with `random`. Pushed commit 1129bbb to use `secrets` instead.

obilodeau: Just to make sure, we don't have a challenge we could use in the TSREQUEST[SPNEGO] from the server? This is when we drop the server's connection, we realize we can't MITM CredSSP so we drop. But we did get that SPNEGO, correct? Doesn't it hold the challenge already?

image

pyrdp/mitm/RDPMITM.py
```diff
@@ -239,11 +239,13 @@ def doClientTls(self):
 
         # Add unknown packet handlers.
         ntlmSSPState = NTLMSSPState()
-        if self.state.ntlmCatch:
-            self.client.segmentation.addObserver(NLAHandler(self.client.tcp, ntlmSSPState, self.getLog("ntlmssp"), True))
-        else:
-            self.client.segmentation.addObserver(NLAHandler(self.server.tcp, ntlmSSPState, self.getLog("ntlmssp")))
-            self.server.segmentation.addObserver(NLAHandler(self.client.tcp, ntlmSSPState, self.getLog("ntlmssp")))
+        self.client.segmentation.addObserver(NLAHandler(self.client.tcp, ntlmSSPState, self.getLog("ntlmssp")))
```
obilodeau: Something here puzzles me. We went from `self.server.tcp`, `self.client.tcp` to `self.client.tcp`, `self.client.tcp`. I'm not sure this is accurate.

pyrdp/security/nla.py
obilodeau: What I don't like about it is that it bundles PyRDP's own state into a class that tracks a pure protocol state. Also, since its impact is tested only on a single `if` and doesn't really need to be passed around, I feel like this is too drastic. That an argument on NLAHandler is enough.

pyrdp/security/nla.py
lubiedo: No, the SPNEGO is sent in an already established NTLMSSP authentication mode. The connection with the server is cut way before any CredSSP is done.

pyrdp/mitm/RDPMITM.py
lubiedo: Sorry for that, copy&paste mistake. Fixed on 367ee65.

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -53,15 +55,19 @@ def parseNTLMSSPNegotiate(self, data: bytes, stream: BytesIO) -> NTLMSSPNegotiat
...
+        targetInfoLen = stream.read(2)
+        targetInfoMaxLen = stream.read(2)
+        targetInfoBufferOffset = stream.read(4)
+        version = stream.read(4)
+        reserved = stream.read(3)
+        revisionCurrent = stream.read(1)
 
-        targetName = self.parseField(data, targetNameFields)
-        targetInfo = self.parseField(data, targetInfoFields)
         return NTLMSSPChallengePDU(serverChallenge)
```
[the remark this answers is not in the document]
lubiedo: Yes and no. `parseField` is returning the fields data inside the PDU where as the change that I've made is getting parsing the data 1:1 in comparison to the `writeNTLMSSPChallenge` function.

pyrdp/security/nla.py
lubiedo: Reverted 20fa506 and using keyword to make it more clear that we are capturing the NTLM hash.

pyrdp/parser/rdp/ntlmssp.py
lubiedo: Should I resolve this?

pyrdp/parser/rdp/ntlmssp.py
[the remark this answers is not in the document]
lubiedo: Also, it is being used by other functions inside this class. Let me know if I can resolve this or needs more attention.

pyrdp/security/nla.py
```diff
@@ -32,16 +34,36 @@ def __init__(self, sink: IntermediateLayer, state: NTLMSSPState, log: logging.Lo
...
 
-        if messageOffset != -1:
-            message: NTLMSSPPDU = self.ntlmSSPParser.parse(data[messageOffset :])
+        if signatureOffset != -1:
+            message: NTLMSSPPDU = self.ntlmSSPParser.parse(data)
             self.ntlmSSPState.setMessage(message)
 
+            if message.messageType == NTLMSSPMessageType.NEGOTIATE_MESSAGE and self.ntlmCapture:
+                randomChallenge = self.getRandChallenge()
+                self.log.info("NTLMSSP Negotiation")
```
obilodeau, line 57: Since it doesn't add value or content from the client, I think this should be changed to a `debug` level message instead of `info`. `info` level message would generate a json event for that log but I don't think it would be valuable.

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -34,11 +37,11 @@ def findMessage(self, data: bytes) -> int:
...
 
     def doParse(self, data: bytes) -> NTLMSSPPDU:
-        stream = BytesIO(data)
+        sigOffset = self.findMessage(data)
+        stream    = BytesIO(data[sigOffset:])
         signature = stream.read(8)
         messageType = Uint32LE.unpack(stream)
-
-        return self.handlers[messageType](data, stream)
+        return self.handlers[messageType](stream.getvalue(), stream)
```
obilodeau, line 44: Before, we would pass a ByteIO stream and all data to the handlers.

Now, no matter the `messageType`, we pass a subset of the data starting with `NTLMSSSP\x00` and a partial stream accordingly.

The behavior might be correct to capture NTLM hashes but it is probably wrong for other packet types.

pyrdp/parser/rdp/ntlmssp.py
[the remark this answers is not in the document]
obilodeau: I'm going to try to improve it myself.

obilodeau (commented):
Another round of review/comments.

obilodeau:
Another important problem showed up. There seems to be a problem if we don't already have a copy of the server's certificate:

```
[2021-11-23 16:50:03,586] - INFO - Ellyn175837 - pyrdp.mitm.connections.x224 - Server requires CredSSP. Closing connection with server and attempting to capture NTLM hashes.
Unhandled Error
Traceback (most recent call last):
  File "/usr/lib/python3.9/asyncio/base_events.py", line 596, in run_forever
    self._run_once()
  File "/usr/lib/python3.9/asyncio/base_events.py", line 1890, in _run_once
    handle._run()
  File "/usr/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/home/olivier/Documents/gosecure/src/pyrdp/venv/lib/python3.9/site-packages/twisted/internet/asyncioreactor.py", line 273, in _onTimer
    self.runUntilCurrent()
--- <exception caught here> ---
  File "/home/olivier/Documents/gosecure/src/pyrdp/venv/lib/python3.9/site-packages/twisted/internet/base.py", line 994, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/home/olivier/Documents/gosecure/src/pyrdp/pyrdp/mitm/RDPMITM.py", line 229, in doClientTls
    privKey, certFile = self.certs.lookup(cert)
builtins.UnboundLocalError: local variable 'cert' referenced before assignment
```

pyrdp/mitm/RDPMITM.py
```diff
@@ -238,6 +239,11 @@ def doClientTls(self):
 
         # Add unknown packet handlers.
```
obilodeau, line 240: I think that comment needs to be revisited. It's just confusing now and since we parse NTLMSSP better we can rewrite it to be better here

obilodeau:
> Another important problem showed up. There seems to be a problem if we don't already have a copy of the server's certificate:

I might have a fix for this in the works.

lubiedo:
Thanks for the reviews @obilodeau! Will work on them ASAP.

lubiedo:
Fix for major issue https://github.com/GoSecure/pyrdp/pull/367#issuecomment-977200529 fixed with commit 527a03b

pyrdp/mitm/X224MITM.py
```diff
@@ -83,6 +83,9 @@ def onConnectionRequest(self, pdu: X224ConnectionRequestPDU):
             # Tell the server we only support the allowed authentication methods.
             chosenProtocols &= self.state.config.authMethods
 
+        if not self.state.ntlmCapture:
```
obilodeau, line 86: This will affect our non-NLA attack scenario. I need to test to confirm but I think it will because this basically says: if we are not capturing NTLM don't perform the downgrade attack. However, when we are facing a non-NLA server we want to perform the downgrade attack.

obilodeau: I can confirm that this is the case. Trying to connect to a system that disables NLA we get:
image

I think we will need to implement the state-tracking route that I was trying yesterday. If the target server enforces NLA and we don't have NLA redirection enabled then we don't downgrade the negotiation. A mix of both our patches from yesterday. You might have another idea so I'll wait until we can talk to confirm.

pyrdp/mitm/X224MITM.py
```diff
@@ -117,24 +121,33 @@ def onConnectionConfirm(self, pdu: X224ConnectionConfirmPDU):
...
+
                 self.onConnectionRequest(self.originalConnectionRequest)
                 return
             else:
                 self.log.info("The server failed the negotiation. Error: %(error)s", {"error": NegotiationFailureCode.getMessage(response.failureCode)})
                 payload = pdu.payload
         else:
-            payload = parser.write(NegotiationResponsePDU(NegotiationType.TYPE_RDP_NEG_RSP, 0x00, response.selectedProtocols))
+            if self.state.ntlmCapture:
+                payload = parser.write(NegotiationResponsePDU(NegotiationType.TYPE_RDP_NEG_RSP, 0x00, NegotiationProtocols.CRED_SSP))
```
obilodeau, line 148: this should be turned into an `elif`

pyrdp/parser/rdp/ntlmssp.py
lubiedo: This was previously done at `pyrdp/security/nla.py` and moved it here.

pyrdp/mitm/X224MITM.py
obilodeau: Fixed by @lubiedo in `52bd5df`

pyrdp/mitm/X224MITM.py
```diff
@@ -83,6 +83,9 @@ def onConnectionRequest(self, pdu: X224ConnectionRequestPDU):
             # Tell the server we only support the allowed authentication methods.
             chosenProtocols &= self.state.config.authMethods
 
+        if self.state.ntlmCapture:
+            chosenProtocols = NegotiationProtocols.SSL | NegotiationProtocols.CRED_SSP
```
obilodeau, line 87: I think we need to describe why we added that here in a comment. Here's a suggestion:

```suggestion
            # If we want to capture the NTLM hash, we need to put back CredSSP in here.
            # If we don't do that we will not get to the state where we can clone the certificate if needed.
            chosenProtocols = NegotiationProtocols.SSL | NegotiationProtocols.CRED_SSP
```

pyrdp/mitm/X224MITM.py
```diff
@@ -117,22 +121,30 @@ def onConnectionConfirm(self, pdu: X224ConnectionConfirmPDU):
...
+                if self.state.canRedirect():
+                    self.log.info("The server forces the use of NLA. Using redirection host: %(redirectionHost)s:%(redirectionPort)d", {
+                        "redirectionHost": self.state.config.redirectionHost,
+                        "redirectionPort": self.state.config.redirectionPort
+                    })
+
+                    # Use redirection host and replay sequence starting from the connection request
+                    self.state.useRedirectionHost()
+                else:
+                    self.log.info("Server requires CredSSP. Reconnecting with server and attempting to capture client's NTLM hashes.")
```
obilodeau, line 138: I think we need a comment and the log needs to be adjusted to be clearer. Suggestion:

```suggestion
                    # If we are not configured to redirect then we should capture the NTLM hash
                    self.log.info("Server requires CredSSP/NLA and we are not configured to support it. Attempting to capture client's NTLM hashes.")
```

I also filed a doc bug #371 that we need to address before release.

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -53,15 +56,19 @@ def parseNTLMSSPNegotiate(self, data: bytes, stream: BytesIO) -> NTLMSSPNegotiat
         return NTLMSSPNegotiatePDU()
 
     def parseNTLMSSPChallenge(self, data: bytes, stream: BytesIO) -> NTLMSSPChallengePDU:
-        targetNameFields = stream.read(8)
+        workstationLen = stream.read(2)
```
obilodeau, line 59: All len fields calls should be converted to:
```
        workstationLen = Uint16LE.unpack(stream)
```

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -86,3 +93,118 @@ def parseNTLMSSPAuthenticate(self, data: bytes, stream: BytesIO) -> NTLMSSPAuthe
...
+
+    def parseNTLMSSPChallengePayload(self, data: bytes, stream: BytesIO, workstationLen: int) -> NTLMSSPChallengePayloadPDU:
+        workstation = stream.read(workstationLen)
+        return NTLMSSPChallengePayloadPDU(workstation)
+
+    def writeNTLMSSPChallenge(self, workstation: str, serverChallenge: bytes) -> bytes:
+        stream = BytesIO()
+        substream = BytesIO()
+
+        workstation = workstation.encode('utf-16le')
```
obilodeau, line 133: We have pack methods for that. It should look like, check `parser/rdp/client_info.py` for inspiration:
```
        Uint16LE.pack(len(extraInfo.clientAddress), stream)
        stream.write(extraInfo.clientAddress)
```

pyrdp/parser/rdp/ntlmssp.py
```diff
@@ -86,3 +93,118 @@ def parseNTLMSSPAuthenticate(self, data: bytes, stream: BytesIO) -> NTLMSSPAuthe
...
+        Uint16LE.pack(pairsLen, substream)
+        Uint16LE.pack(pairsLen, substream)
+        Uint32LE.pack(NTLMSSPChallengeType.WORKSTATION_BUFFER_OFFSET + nameLen, substream)
+        Uint8.pack(NTLMSSPChallengeVersion.NEG_PROD_MAJOR_VERSION_HIGH, substream)
+        Uint8.pack(NTLMSSPChallengeVersion.NEG_PROD_MINOR_VERSION_LOW, substream)
+        Uint16LE.pack(NTLMSSPChallengeVersion.NEG_PROD_VERSION_BUILT, substream)
+        Uint8.pack(0, substream)
+        Uint8.pack(0, substream)
+        Uint8.pack(0, substream)
+        Uint8.pack(NTLMSSPChallengeVersion.NEG_NTLM_REVISION_CURRENT, substream)
```
obilodeau, line 154: Where is that format documented? pairsLen, nameLen is written so often, I think there's something here that is unclear.

Could we do something similar to this?
```python
        domainLength = len(domain) - nullByteCount * unicodeMultiplier
        usernameLength = len(username) - nullByteCount * unicodeMultiplier
        passwordLength = len(password) - nullByteCount * unicodeMultiplier
        alternateShellLength = len(alternateShell) - nullByteCount * unicodeMultiplier
        workingDirLength = len(workingDir) - nullByteCount * unicodeMultiplier

        stream.write(Uint16LE.pack(domainLength))
        stream.write(Uint16LE.pack(usernameLength))
        stream.write(Uint16LE.pack(passwordLength))
        stream.write(Uint16LE.pack(alternateShellLength))
        stream.write(Uint16LE.pack(workingDirLength))
        stream.write(domain)
        stream.write(username)
        stream.write(password)
        stream.write(alternateShell)
        stream.write(workingDir)
```

obilodeau (commented):
Very minor suggestions/comments

pyrdp/parser/rdp/ntlmssp.py
lubiedo: The documentation can be found in the [PDU](https://github.com/lubiedo/pyrdp/blob/issue-358/pyrdp/pdu/rdp/ntlmssp.py#L30-L37) but here's the link: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/801a4681-8809-4be9-ab0d-61dcfe762786

This is an specific struct that I'm trying to follow verbatim for better debugging and clarity. The reason why `nameLen` and `pairsLen` is written twice is because the first write refers to the length and the second to the max length, which are same in this case. This is an excerpt of the documentation, in this case for `nameLen`:

```
§ TargetNameLen (2 bytes): A 16-bit unsigned integer that defines the size, in bytes, of TargetName in Payload.
§ TargetNameMaxLen (2 bytes): A 16-bit unsigned integer that SHOULD be set to the value of TargetNameLen and MUST be ignored on receipt.
```

pyrdp/parser/rdp/ntlmssp.py
obilodeau: Forget it, this is not related to writing to the stream so encoding here is fine as a prior step.

pyrdp/parser/rdp/ntlmssp.py
obilodeau: This length, maxLength but is the same deal is so confusing. Not your fault, that's on Microsoft. Add that link to the method documentation and it's fine.

obilodeau (approved):
This PR sparks joy! :star2:

PR #366, opened 2021-10-26
fix(331): fix json conversion

alxbl:
This PR addresses several issues with the pyrdp-convert refactor:

- Missing scapy imports leading to exceptions when converting
- Bad `__iter__` implementation for `ExportedPDUStream` resulting in improper cleanup
- Missing handler cleanup call in `PCAPConverter`

Among other things, it closes #331.

Tested on both encrypted pcaps and ExportedPDU pcaps.

PR #368, opened 2021-10-27
feat(#313): Add support for conversion of multiple exported sessions.

### LINKED ISSUE #313: Parsing a decrypted PDU-L7 Pcap with multiple sessions does not work
I am able to reliably reproduce a bug where a Pcap with exported PDUs containing more than one session of the same source and destination IPs is getting mixed up and doesn't convert to the correct replay file or mp4.

We should probably cleanly separate streams in the PDU parsing code. Not sure how to do that with PDU level packets (is the information available in the wireshark trace?).

Workaround: walking the pcap with filters like `tcp.stream eq <1, 2, .. N>` and exporting one PDU-L7 pcap per TCP streams works.

alxbl:
This pull request adds the ability to convert multiple sessions inside of an exported PDU PCAP. It fixes #313.

This is done by parsing the EXPORTED_PDU header (in a very ad-hoc way, right now) with a custom `Exported` layer to retrieve the `ip.src`, `tcp.sport`, `ip.dst`, and `tcp.dport` tags in the header and modifying the `tcp_both` session extraction code to use the `Exported` layer.

I say the layer is ad-hoc because `EXPORTED_PDU`'s tags are an arbitrary list of tags with the following form:

```
tagIdentifier (2B)
tagLength (2B)
tagBytes (<tagLength>B)
[.... repeat until tagEnd ...]
tagEnd ("\x00\x00")
tagLength ("\x00\x00")
```

But right now what I'm doing is hardcoding the list of tags that I've seen in typical exported RDP L7 captures, so it might fail for non-standard exports.

We can fix that when/if it happens. In general if it fails, it's probably because the pcap wasn't exported properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,41 +76,9 @@ jobs:
         working-directory: ./
         run: coverage run --append test/test_mitm_initialization.py dummy_value
 
-      - name: pyrdp-player.py read a replay in headless mode test
+      - name: Running pyrdp-player and pyrdp-convert integration tests with verifications
         working-directory: ./
-        run: coverage run --append bin/pyrdp-player.py --headless test/files/test_session.replay
-
-      - name: pyrdp-convert.py to MP4
-        working-directory: ./
-        run: coverage run --append bin/pyrdp-convert.py test/files/test_convert.pyrdp -f mp4
-
-      - name: Verify the MP4 file
-        working-directory: ./
-        run: file test_convert.mp4 | grep "MP4 Base Media"
-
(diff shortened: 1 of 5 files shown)


obilodeau:
Not much more luck here. I'll investigate the JSON output deeper then.

obilodeau:
Just pushed a rebase on master. I manually handled conflicts but the new `InetAddress` code is still not using `Exported` since I fixed similar problems in my own fixes on top of the #366 PR. More stuff is coming.

obilodeau:
Another interactive rebase to make everything neat and avoid a squash commit

PR #369, opened 2021-10-27
fix: Better success messages for pyrdp-convert

alxbl:
While making my other PRs, I noticed that the pyrdp-convert success message was not very consistent, so I wrote a bit of code to make it consistent.

It will print the filename when outputting to the same folder, or the absolute file path if a prefix is specified. For multiple streams, each file will be printed after its conversion.

The way this was done was by exposing the output filename to the `RDPReplayer` via a property that gets it from the transport layer (either a `ConversionLayer` or a `FileLayer`)

We could make this a bit sturdier by having a common base class for both `ConversionLayer` and `FileLayer` that enforces the presence of a `filename` member, but for now both have it, so it works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,10 @@ jobs:
         working-directory: ./
         run: coverage run --append bin/pyrdp-convert.py test/files/test_session.pcap -f replay
 
+      - name: Verify that the replay file exists
+        working-directory: ./
+        run: file -E "20200319000716_192.168.38.1\:20989-192.168.38.1\:3389.pyrdp"
+
       - name: Run unit tests
         working-directory: ./
         run: coverage run --append -m unittest discover -v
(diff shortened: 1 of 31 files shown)


obilodeau:
Tested this but I stumble upon the same problems as #366. Moving to the next one.

obilodeau:
Now that #366 is merged. I started to look into this one tonight. Unfortunately, this PR doesn't work with the replay handler because, in it, it says:

https://github.com/GoSecure/pyrdp/blob/44e94a99906350cc897db1bd403698eaad0b72eb/pyrdp/convert/utils.py#L84

This should have been caught by tests. I'll rework them to make sure it is.

obilodeau:
The error should trigger with tests now. I added nice little JSON tests while we are at it.

The history is a bit of a mess because I pulled this PR in my master back in November so when I pushed back a lot of things went up with it. That said, the GitHub diff is clean so I'm leaving this that way. In any case, I'll probably squash-merge.

obilodeau:
Ok, so all errors were swallowed so adding tests didn't help. Now I'm adjusting exit codes but this will make existing tests fail I think. I'll see what I do about that tomorrow. Zzz

obilodeau:
Ohh, actually this seems to be exactly what I wanted!

image

Now time to fix that missing `handler` problem and it is ready for merge.

PR #386, opened 2022-01-18
Pre-release dependency updates

obilodeau:
Local tests passed, let's see CI


obilodeau:
Added this security update as well: https://github.com/GoSecure/pyrdp/pull/391

PR #403, opened 2022-07-08
Enable play/pause replay from keyboard

lubiedo:
Simple change to play or pause reproduction of a replay via the space bar.


pyrdp/player/RDPMITMWidget.py
```diff
@@ -197,4 +197,4 @@ def sendCurrentScreen(self):
         width = self._buffer.width()
         height = self._buffer.height()
         pdu = PlayerBitmapPDU(self.layer.getCurrentTimeStamp(), width, height, self._buffer.bits())
-        self.layer.sendPDU(pdu)
\ No newline at end of file
+        self.layer.sendPDU(pdu)
```
obilodeau, line 200: Try to avoid unrelated changes, use `git add -p` (or the GUI equivalent) instead of `git add` to select what you commit.

Not for this PR but for the next one

pyrdp/ui/PlayPauseButton.py
```diff
@@ -32,4 +33,4 @@ def setPlaying(self, playing):
             self.setIcon(QIcon.fromTheme("media-playback-pause"))
         else:
             self.setText(self.playText)
-            self.setIcon(QIcon.fromTheme("media-playback-start"))
\ No newline at end of file
+            self.setIcon(QIcon.fromTheme("media-playback-start"))
```
obilodeau, line 36: Try to avoid unrelated changes, use `git add -p` (or the GUI equivalent) instead of `git add` to select what you commit.

Not for this PR but for the next one

pyrdp/ui/PlayPauseButton.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, parent: QWidget = None, playText = "Play", pauseText = "Pause
         self.pauseText = pauseText
         self.playing = False
         self.setPlaying(self.playing)
+        self.setFixedWidth(100)
```
obilodeau, line 23: Why this change?

obilodeau (changes requested):
It's a great UX improvement! I assume you tested it to work, right?

Can you add a CHANGELOG entry? I'll merge right after.

pyrdp/ui/PlayPauseButton.py
lubiedo: Yeah, will do.

pyrdp/player/RDPMITMWidget.py
lubiedo: Yeah, will do.

pyrdp/ui/PlayPauseButton.py
lubiedo: This gets a constant fixed width to the play/pause button as, before this, when the label toggled from `Play` to `Pause` the width changed. It is more of a design change.

lubiedo:
Yes, tested it against replays and it worked without any issue. Also, added the CHANGELOG entry on [465291b](https://github.com/GoSecure/pyrdp/pull/403/commits/465291bc28817f4f8e55e5e9f6506a0ac3dad92d).

obilodeau (approved):
Good to go

PR #424, opened 2022-11-25
nla redirection: use certificate of original server

spameier:
fixes #423

PR #426, opened 2022-12-13
WIP: Fake RDP server when NLA is enforced

spameier:
As part of a project at school, I extended the existing redirection functionality with a locally hosted RDP server. It works as follows:
1. A local RDP server is with a fake login screen and no authentication is launched (A Tkinter GUI with freerdp-shadow-cli inside Xephyr).
2. Once a user tries to connect to a NLA enforcing RDP server, the traffic is redirected to the fake login screen.
3. Once the user enters credentials, a rdp client (xfreerdp) is launched, replacing the fake login screen.
4. The user now sees the remote server through fake RDP server.

So far it's work in progress and has a lot of rough edges, but feedback would be very much appreciated.


pyrdp/mitm/RDPMITM.py
```diff
@@ -218,7 +220,20 @@ async def connectToServer(self):
...
+                    "host": self.state.config.targetHost,
+                    "port": self.state.config.targetPort,
+                },
+            )
+            pem = ssl.get_server_certificate(
+                (self.state.config.targetHost, self.state.config.targetPort)
+            )
+            cert = crypto.load_certificate(crypto.FILETYPE_PEM, pem)
+        else:
+            cert = self.server.tcp.transport.getPeerCertificate()
```
obilodeau, line 236: There are issues with this code that are being looked at in #424

pyrdp/mitm/FakeServer.py
```diff
@@ -0,0 +1,274 @@
+#
+# This file is part of the PyRDP project.
+# Copyright (C) 2022
+# Licensed under the GPLv3 or later.
+#
+import os, random, shutil, socket, subprocess, threading, time
+from tkinter import *
+from PIL import Image, ImageTk
+from pyvirtualdisplay import Display
+from pyrdp.logging import log  # TODO: this logs to GLOBAL and not to the session, setup proper logging
```
obilodeau, line 10: imports should be organized PEP8 style: stdlib first then external dependencies then local

obilodeau, line 10: Let's make sure we do per-session logging with JSON events being generated properly as well.

pyrdp/mitm/FakeServer.py
```diff
@@ -0,0 +1,274 @@
...
+        self.entry_username.place(
+            relx=0.5, rely=0.61, anchor=CENTER, height=40, width=290
+        )
+        self.entry_username.focus()
+
+        # password entry
+        self.entry_password = Entry(
+            self.root,
+            show="•",
+            font=("Segoe UI", 20),
```
obilodeau, line 86: Does this add special dependencies?

pyrdp/mitm/FakeServer.py
```diff
@@ -0,0 +1,274 @@
...
+    def run(self):
+        self.fakeLoginScreen = FakeLoginScreen()
+        self.fakeLoginScreen.show()
+        username = self.fakeLoginScreen.username
+        password = self.fakeLoginScreen.password
+        self.fakeLoginScreen = None
+
+        rdp_client_cmd = [
+            shutil.which("xfreerdp"),
+            "/v:192.168.251.12",  # TODO make dynamic,
```
obilodeau, line 231: hardcoded

obilodeau (changes requested):
This is very interesting and in a direction I had never thought of. Great work! I would love to see it in action if you can provide a recorded session.

Missing:
* Impact on dependencies outside of python (X11, TK, Fonts, freerdp-shadow-cli, xfreerdp)
* Documentation
* Handle TODOs and FIXMEs

spameier:
> I would love to see it in action if you can provide a recorded session.
Here's a quick demo:
- Client: https://pixeldrain.com/u/K5m5dJwi
- Kali: https://pixeldrain.com/u/c7U5jUbn

pyrdp/mitm/FakeServer.py
spameier: done

pyrdp/mitm/FakeServer.py
spameier: I *think* that's solved.

spameier: Do you see a better way to add per session logging, other that to introduce logging to `state.py`? If yes, that might make more sense. I'm also not entirely sure if we should import the fake server in `state.py` as it is right now or better in `RDPMITM.py`

> with JSON events being generated properly as well.

Do you mean that the logging messages should be parametrized? If yes, I think that should be solved now.

pyrdp/mitm/FakeServer.py
spameier: Good point! I'm not sure honestly, it's certainly not on my Kali VM and I never exactly looked at it. Also I'm not a fonts expert :smile:. I removed it for now, since it doesn't add much.

spameier: I'm switching to DejaVu Sans but that's not available on Windows I think. I also didn't find a way to only specify a font family (e. g. Sans Serif)..

pyrdp/mitm/FakeServer.py
obilodeau: Yes, this looks better now. Thanks!

spameier:
Thanks for your feedback!

> I think the fake server should be headless with the option of using the pyrdp-player to listen to the connection just like with pyrdp when it is operating regularly.

Yes, of course. Using `Xephyr` with the GUI is useful while testing and for demonstration purposes. Luckily `pyvirtualdisplay` already supports the headless `Xvfb` as a backend. I don't think it will give much trouble to change that and will look into it in the future.

The CI is currently failing because I added the `getLog` callback to `RDPMITMState`. I think that should be solved differently anyway.

spameier:
Making the server headless is as easy as this. (One drawback: the background after the login screen and before xfreerdp loads is black):
```diff
diff --git a/pyrdp/mitm/FakeServer.py b/pyrdp/mitm/FakeServer.py
index 96a74ce..89f2681 100644
--- a/pyrdp/mitm/FakeServer.py
+++ b/pyrdp/mitm/FakeServer.py
@@ -173,11 +173,7 @@ class FakeServer(threading.Thread):
         self._launch_rdp_server()
 
     def _launch_display(self, width=1920, height=1080):
-        self.display = Display(
-            backend="xephyr",
-            size=(width, height),
-            extra_args=["-no-host-grab", "-noreset"],
-        )  # noreset for xsetroot required
+        self.display = Display(backend="xvfb", size=(width, height))
         self.display.start()
         self.display.env()
         # set background to windows blue
```

PR #433, opened 2023-04-03
Update rdp.py

plonxyz:
adding RDP10_11 due to error builtins.ValueError: 524304 is not a valid RDPVersion when connecting via win11 RDP


obilodeau:
Thanks for your contribution!

Issue #436, opened 2023-06-15
PySide2 incompatibility with Python 3.11


obilodeau:
PySide2 is explicitly incompatible with Python 3.11. When doing a `pip install full` of PyRDP on a 3.11 install:

```
ERROR: Could not find a version that satisfies the requirement PySide2<6,>=5.12; extra == "full" (from pyrdp[full]) (from versions: none)
ERROR: No matching distribution found for PySide2<6,>=5.12; extra == "full"
```

From https://pypi.org/project/PySide2/:

![image](https://github.com/GoSecure/pyrdp/assets/546325/04926588-8711-4c73-b8d1-826196b252da)

QT (maintainers of PySide) are saying that 5.15.11 is now LTS which is no longer available in built form... See https://forum.qt.io/topic/144843/pyside2-shiboken2-support-for-python-3-11.

We will need to consider migrating to QT6 eventually...

As a workaround, make sure to install a 3.10.x Python.

obilodeau:
Appears to only affect Windows since I installed PyRDP from source with python 3.11 today

TiddlyWiddly:
Affects newest version of Kali on M1 too, unfortunately...

obilodeau:
> Affects newest version of Kali on M1 too, unfortunately...

To be precise, it affects all versions of Kali M1 since PySide2 is not built for the M1 architecture: https://forum.qt.io/topic/124289/mac-m1-pyside-2-6-can-t-install-lib/12.

This would be a different issue but with the same fix.

obilodeau:
> Affects newest version of Kali on M1 too, unfortunately...

@TiddlyWiddly you can try Kali on M1 again on master. It should work now.

Issue #316, opened 2021-06-02
PyRDP running error on Windows 10


yraiesets:
Both pyrdp-mitm.py and pyrdp-player.py trigger a TypeError on execution.

Platform : Windows 10.
Python Version : 3.9.5

`Traceback (most recent call last):
  File "C:\Users\admin\Desktop\pyrdp\bin\pyrdp-mitm.py", line 15, in <module>
    asyncioreactor.install(asyncio.get_event_loop())
  File "C:\Users\admin\AppData\Local\Programs\Python\Python39\lib\site-packages\twisted\internet\asyncioreactor.py", line 307, in install
    reactor = AsyncioSelectorReactor(eventloop)
  File "C:\Users\admin\AppData\Local\Programs\Python\Python39\lib\site-packages\twisted\internet\asyncioreactor.py", line 60, in __init__
    raise TypeError(
TypeError: SelectorEventLoop required, instead got: <ProactorEventLoop running=False closed=False debug=False>`

obilodeau:
Something related. I think most of the integration tests on Windows are worthless. Even if the return code is 0 check the individual outputs.

obilodeau:
Tonight on Windows 10 with Python 3.10.10:

```
python venv\Scripts\pyrdp-player.py
Traceback (most recent call last):
  File "C:\[...]\tools\pyrdp\venv\Scripts\pyrdp-player.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "C:\[...]\tools\pyrdp\bin\pyrdp-player.py", line 11, in <module>
    asyncioreactor.install(asyncio.new_event_loop())
  File "C:\[...]\tools\pyrdp\venv\lib\site-packages\twisted\internet\asyncioreactor.py", line 306, in install
    reactor = AsyncioSelectorReactor(eventloop)
  File "C:\[...]\tools\pyrdp\venv\lib\site-packages\twisted\internet\asyncioreactor.py", line 61, in __init__
    raise TypeError(
TypeError: ProactorEventLoop is not supported, got: <ProactorEventLoop running=False closed=False debug=False>
```

obilodeau:
The PyRDP Player works if you add this line below the `import asyncio` call at the top of the `bin/pyrdp-player.py` file:

```python
asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```

Not sure what the real fix will be at this point but an OS-specific reactor could do it.

[...]

PR #439, opened 2023-06-27
Test Python LTS and Python latest-1 on Windows with CI

obilodeau:


obilodeau:
Build fails due to #436, I'll revert to 3.10 on Windows as the "latest"

obilodeau:
And now it seems to be due to #316

PR #444, opened 2023-07-20
fix pyrdp running error on windows10+

Mas0nShi:
Hi, #316 still exist in [pyrdp-mitm.py](https://github.com/GoSecure/pyrdp/blob/master/bin/pyrdp-mitm.py) and [pyrdp-clonecert.py](https://github.com/GoSecure/pyrdp/blob/master/bin/pyrdp-clonecert.py), i've fixed it now (same solution as #439) :).


obilodeau:
Thanks for your contribution!

Issue #445, opened 2023-07-28
pyrdp player logs are different with/without headless


i-am-shodan:
I'm using the latest version of pyrdp

When I run:
`venv/bin/pyrdp-player.py /mnt/d/rdpdebug/19700101000000_191.101.157.12%3A50217-10.0.13.101%3A3389.pyrdp`

I get this log at the bottom of the screen

```
--------------------
HOST: iPhone

--------------------

--------------------
USERNAME: Administrator
PASSWORD: MYPASSWORD
DOMAIN: 

--------------------
notepad
<Return pressed>
<Return released>hellhello world!
<Connection closed>
```

But when I run:

`venv/bin/pyrdp-player.py --headless /mnt/d/rdpdebug/19700101000000_191.101.157.12%3A50217-10.0.13.101%3A3389.pyrdp`

I get either this debug out file

```
[2023-07-28 14:49:28,808] - INFO - pyrdp - Starting PyRDP Player in headless mode.
```
or this from stdout

```
== REPLAY FILE: 19700101000000_191.101.157.12%3A50217-10.0.13.101%3A3389.pyrdp                                                                                                                                                                                                                                                                                                                    --------------------                                                                                                                                                                             HOST: iPhone                                                                                                                                                                                                                                                                                                                                                                                      --------------------                                                                                                                                                                                                                                                                                                                                                                              -- END --------------------------------
```

They should be the same

i-am-shodan:
Seems like Replay is only returning 1 event, the first one

obilodeau:
I can't reproduce this from my setup and @robeving's fix puzzles me so I want to test it out.

Can you provide the `.pyrdp` file? How was it created? With `pyrdp-mitm` or with `pyrdp-convert`?

i-am-shodan:
I can provide you an original decrypted PCAP if you like and the command lines I use with pyrdp-convert to get the result.

Basically timestamp is always 0 when I do:
Enc pcap -> decrypt with tshark -> convert with pyrdp-convert -> replay to get logfile

The only thing unusual in my set up is that the pcap is constructed with a library coming out of our custom RDP load balancer. That being said it looks valid in wireshark and tshark is producing the decrypted pcap.

obilodeau:
Yes, getting a decrypted pcap would be great. Attach it here or send it to my username at gosecure.net.

obilodeau:
[...]

PR #446, opened 2023-07-28
Fixed bug with PDU handling in headless mode

robeving:
This fixes a bug with headless replay of files, specifically those created by pyrdp-convert.py which seems to set the timestamp to 0. Using the timestamp is probably bad anyway due to the risk of collision.


robeving:
Fixes https://github.com/GoSecure/pyrdp/issues/445

Issue #351, opened 2021-08-30
Broken player - qt.qpa.pluging: Could not load the QT platform plugin "xcb"


Norup123:
On a fresh ubuntu 20.04.3 install i have tried installing the pyrdp. I followed the install guide with pip, and the mitm is working fine. When trying to run the player however i am presented with the following issue

```
~/pyrdp$ pyrdp-player.py -p 3000
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

Aborted (core dumped)

```

Any idea what could be the issue here? I assume something broke during the installation, but i tried running the command
`pip3 install -U -e '.[full]'` again, but get the same error

Norup123:
Ran `sudo apt-get install '^libxcb.*-dev'` outside the venv and now it works

obilodeau:
What is your desktop environment? I have an 20.04 LTS installed but w/o a GUI for now. I will try to add the correct dependency but I don't want to add `'^libxcb.*-dev'` as I think it will increase our docker image significantly and only a small part of it is required.

Norup123:
![image](https://user-images.githubusercontent.com/54578187/131483266-0d211f56-a19e-4358-a09e-7079fc0d7eca.png)

obilodeau:
I managed to reproduce the issue in a VM
![2021-08-31_10-22 issue 351](https://user-images.githubusercontent.com/546325/131510039-172aa3f6-8d36-4cdd-8024-9f5328ec36cd.png)

obilodeau:
Running `export QT_DEBUG_PLUGINS=1` then running `pyrdp-player.py` again gave a good amount of helpful output including:
```
QFactoryLoader::QFactoryLoader() checking directory path "/usr/bin/platforms" ...
Cannot load library /home/vagrant/pyrdp/venv/lib/python3.8/site-packages/PySide2/Qt/plugins/platforms/libqxcb.so: (libxcb-xinerama.so.0: cannot open shared object file: No such file or directory)
[...]

PR #448, opened 2023-12-02
Add missing libqxcb dependencies

kszafran:
I run into an issue similar to https://github.com/GoSecure/pyrdp/issues/351 running pyrdp-convert.py (or pyrdp-player.py) in docker (both the latest image from dockerhub, as well as an image built locally from the Dockerfile):
```
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.

Aborted (core dumped)
```
Running with `QT_DEBUG_PLUGINS` I found out that libqxcb failed to load:
```
Cannot load library /opt/venv/lib/python3.8/site-packages/PySide2/Qt/plugins/platforms/libqxcb.so: (libxcb-icccm.so.4: cannot open shared object file: No such file or directory)
``` 
Digging further with `ldd` I identified a couple of missing libraries:
```
ldd /opt/venv/lib/python3.8/site-packages/PySide2/Qt/plugins/platforms/libqxcb.so | grep 'not found'
        libxcb-icccm.so.4 => not found
        libxcb-image.so.0 => not found
        libxcb-util.so.1 => not found
        libxcb-keysyms.so.1 => not found
        libxcb-randr.so.0 => not found
        libxcb-render-util.so.0 => not found
```

Installing these libraries with apt-get in the Dockerfile fixed the issue and I was able to produce an mp4 output. Perhaps something has changed in Ubuntu repositories.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends python3 \
         python3-distutils \
         # GUI and notifications stuff
         libgl1-mesa-glx libxcb-xinerama0 \
+        libxcb-icccm4 libxcb-image0 libxcb-util1 libxcb-keysyms1 \
+        libxcb-randr0 libxcb-render-util0 \
         libxkbcommon-x11-0 \
         libnotify-bin \
         # Runtime requirement for PyAV (pyrdp-convert to MP4)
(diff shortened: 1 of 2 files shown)


obilodeau:
Thanks! Will need to adjust installation instructions in the README as well.

kszafran:
@obilodeau Thank you, I updated the README.

obilodeau:
Thanks for the update!

PR #464, opened 2024-01-15
Set a default value for clipboard clipDataId

luciaprime54:
xfreeRDP (and maybe other non-Windows clients?) don't send a `clipDataId`, which makes the code crash when trying to extract a file being copied over RDP.

Currently we don't use `clipDataId` for anything, so we can safely just set it to 0.


obilodeau:
Read the spec, this sounds correct indeed! Thanks!

obilodeau:
Fix confirmed to work

PR #468, opened 2024-01-19
added --version (or -V) to all CLI tools

obilodeau:


obilodeau:
Let's hold onto this until we support Python >= 3.8 only

PR #471, opened 2024-02-05
Added a catch-all for NTSTATUS for error codes we don't care about

luciaprime54:
Some Clients can return other Error codes that we can safely ignore, such as the MacOS Windows RDP Client when transferring multiple files.

Rather than hardcode 1000s of possible Status codes Just handle the ones we need to and create a catch-all for the ones we don't care about so we don't throw an exception and crash unnecessarily.

Unfortunately I don't have the specific error code I saw on Mac (it was a while ago, and it wasn't every time), but this solution would hopefully also help if other clients do the same thing.


obilodeau (approved):
Looking good. Thank you!

Issue #475, opened 2024-05-23
Add support for RDP version 10.12


kobbi-strike:
Error:
```
[2024-05-23 13:19:22,077] - ERROR - wizardly_mirzakhani_5755447 - pyrdp.mitm.connections.tcp - 524305 is not a valid RDPVersion
Traceback (most recent call last):
....
    version = RDPVersion(Uint32LE.unpack(stream))
  File "/usr/lib/python3.10/enum.py", line 385, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib/python3.10/enum.py", line 710, in __new__
    raise ve_exc
ValueError: 524305 is not a valid RDPVersion
```

kobbi-strike:
Tried adding ;
    RDP10_12 = 0x80011 to enum/rpd.py

Without any luck.

energister:
@kobbi-strike , why did you close the issue? It's still relevant

PR #477, opened 2024-06-12
Handle missing RDPVersion numbers thrown by some RDP clients

luciaprime54:
A recent update of some clients such as Remmina on Linux recently updated the RDP version code to `0x80011` and threw an exception crashing unnecessarily.

Rather than waiting to hardcode the next version number, handle the ones we need to and create a catch-all for the ones we haven't seen.

PR also adds the latest version I saw in the wild: `0x80011`

### DIFF
--- a/pyrdp/enum/rdp.py
+++ b/pyrdp/enum/rdp.py
@@ -241,6 +241,17 @@ class ConnectionDataType(IntEnum):
 
 
 class RDPVersion(IntEnum):
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


energister:
This PR is attempt to address #475.

Unfortunately, it doesn't help me: I'm getting errors and disconnection on the Windows 11 target host (see Event Viewer) and connection timeout at the Windows 11 mstsc client on the other side.
image

Issue #485, opened 2025-02-18
Build fails on Kali rolling 2024.4 : ERROR: Failed building wheel for av


AppSecAmael:
### Context
I have automated the installation of pyrdp on a Kali linux distribution supported by Vagrant. On previous version installation was fine, but now that I switched to 2024.4 image of kali-rolling (https://portal.cloud.hashicorp.com/vagrant/discover/kalilinux/rolling) it fails.

### What I have done : 
1) Installation of PyRDP prerequisites using apt : 
      - libxcb-cursor0
      - python3
      - python3-pip
      - python3-venv
      - build-essential
      - python3-dev
      - git
      - openssl
      - libnotify-bin
      - libxkbcommon-x11-0
      - libxcb-xinerama0
      - libxcb-icccm4
      - libxcb-image0
      - libxcb-util1
      - libxcb-keysyms1
      - libxcb-randr0
      - libxcb-render-util0
      - libavformat-dev
      - libavcodec-dev
      - libavdevice-dev
      - libavutil-dev
      - libswscale-dev
      - libswresample-dev
      - libavfilter-dev
      - ffmpeg
      - python3-av

2)  Read-write git checkout of PyRDP : https://github.com/gosecure/pyrdp.git (HEAD)
3) Install python setuptools in a virtualenv  (pip, wheel, setuptools)
4) Install PyRDP itself : ` ./pyrdp/venv/bin/pip3 install --upgrade --editable .[full]` 

### Issue 

```

DEPRECATION: Loading egg at /home/user/workspace/tools/pyrdp/venv/lib/python3.13/site-packages/cryptography-41.0.7-py3.13-linux-x86_64.egg is deprecated. pip 25.1 will enforce this behaviour change. A possible replacement is to use pip for package installation. Discussion can be found at https://github.com/pypa/pip/issues/12330
DEPRECATION: Loading egg at /home/user/workspace/tools/pyrdp/venv/lib/python3.13/site-packages/progressbar2-4.5.0-py3.13.egg is deprecated. pip 25.1 will enforce this behaviour change. A possible replacement is to use pip for package installation. Discussion can be found at https://github.com/pypa/pip/issues/12330
[...]

PR #487, opened 2025-02-23
Fixes for Python 3.13 compatibility

### LINKED ISSUE #486: Python 3.13 incompatibility due to usage of `typing.io`
When using pyrdp with Python 3.13, it will throw errors like `ImportError: cannot import name 'io' from 'typing'. Did you mean: 'IO'?`.
This is because the `typing` module [changed](https://docs.python.org/3/library/typing.html#abcs-for-working-with-io) and the deprecated interface was removed in 3.13 .

(I could create a PR to fix compatibility but unfortunately I do not see any recent maintenance activity here.)

exploide:
Prior to this PR, pyrdp is not compatible with Python 3.13.

- changed deprecated import to `typing.BinaryIO` (which is backwards compatible even with Python 3.7)
- fixed invalid escape sequence warnings by using raw-strings

fixes #486

@Res260 could you take a look?


Res260:
Hey, could you please modify 
- `.github/workflows/ci.yml` line 39 and 109 to use python 3.13 as well as 3.12
- `.github/workflows/wheels.yml` to use upload-artifacts@v4 instead of @v3 (this is unrelated to your PR, but makes two CI checks fail)

exploide:
Thank you very much for taking care of this PR @Res260 !

> .github/workflows/ci.yml line 39 and 109 to use python 3.13 as well as 3.12

Done, though I don't know why testing against 3.12 should be kept. But I did the change anyway.

> .github/workflows/wheels.yml to use upload-artifacts@v4 instead of @V3 (this is unrelated to your PR, but makes two CI checks fail)

Done, but I'm not sure if it will immediately succeed since the workflow does not run automatically. According to the [migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) changes might be required. Not sure if they are necessary here because the artifact is not named explicitly.

exploide:
Indeed, it was necessary to update the `upload-artifacts@v4` usage as well, now it seems to work.

Besides, it was necessary to update the dependency on `av`, since older versions did no longer build on recent systems. See also #485.

Almost everything seems to work now. I'm not sure why the macOS build isn't running.

PR #489, opened 2025-03-12
Feature - added absolute date and time to player

jmvermeulen:
The absolute time of played frame shown next to the process-bar.
image

Tested with a stream with a length more then a few days, works perfectly.

It's based on the first epoch timestamp of the replay, then the relative time position is added during the playback.


obilodeau:
Thanks! This looks good and useful. Running tests.

Issue #294, opened 2021-02-04
Support for xrdp backend


khast3x:
Hello,

Does PyRDP support using the mitm feature in front of a [xrdp](https://github.com/neutrinolabs/xrdp) server? If not, this could help deploy honeypots with a small alpine/ubuntu container acting as RDP backend.
This is the current output when running as a mitm for a ubuntu+xrdp container (it is a bit verbose, sorry):

```shell
user@dockervps:~# docker run -d --name uxrdp --hostname terminalserver --shm-size 1g danielguerra/ubuntu-xrdp
user@dockervps:~# docker run -p 3389:3389 gosecure/pyrdp pyrdp-mitm.py 172.17.0.2
[2021-02-04 18:53:46,507] - INFO - GLOBAL - pyrdp.mitm - Target: 172.17.0.2:3389
[2021-02-04 18:53:46,507] - INFO - GLOBAL - pyrdp.mitm - Output directory: /home/pyrdp/pyrdp_output
[2021-02-04 18:53:46,509] - INFO - GLOBAL - pyrdp - MITM Server listening on port 3389
[2021-02-04 18:54:16,729] - INFO - Jeanne859549 - pyrdp.mitm.connections.tcp - New client connected from xx.xx.xx.22
[2021-02-04 18:54:16,730] - INFO - Jeanne859549 - pyrdp.mitm.connections.x224 - No cookie for this connection
[2021-02-04 18:54:16,733] - INFO - Jeanne859549 - pyrdp.mitm.connections.tcp - Server connected
[2021-02-04 18:54:17,856] - INFO - Jeanne859549 - pyrdp.mitm.connections.cert - Cloned server certificate to pyrdp_output/certs/Terminalserver.crt
CLIENT_RANDOM 605840d49e2b351091cb3c83a154c7923c09e1a634c24958c494429050124282 c0898b9270fb5aa7483ac14c8deb1fdfa3c89921419396cbf364f08b8b8f81f7ad34098bd28257bf611a8a11546a03aa
[2021-02-04 18:54:20,520] - INFO - Jeanne859549 - pyrdp.mitm.connections.mcs - Client hostname shinyboi
CLIENT_RANDOM 13e4e08473315c9113328b458a64b9ca6b380172d89e06ffdae5fad186ab2068 aab6858872e3782f775abaae75feae9e0d20854de346676dc44ffa3f7c17b2456a0350a184f03f03bbedc2cacdb1eeee
[...]

PR #491, opened 2025-04-08
Fix: #294 Support XRDP backend

yaleman:
This takes the comments from #294 and makes it work as far as I can tell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,36 +10,36 @@ PyRDP is a Python Remote Desktop Protocol (RDP) Monster-in-the-Middle (MITM) too
 ![PyRDP Logo](https://raw.githubusercontent.com/GoSecure/pyrdp/main/docs/pyrdp-logo.svg?sanitize=true)
 
 It features a few tools:
+
 - RDP Monster-in-the-Middle
-    - Logs plaintext credentials or NetNTLM hashes used when connecting
-    - Steals data copied to the clipboard
-    - Saves a copy of the files transferred over the network
-    - Crawls shared drives in the background and saves them locally
-    - Saves replays of connections so you can look at them later
-    - Runs console commands or PowerShell payloads automatically on new connections
+  - Logs plaintext credentials or NetNTLM hashes used when connecting
+  - Steals data copied to the clipboard
+  - Saves a copy of the files transferred over the network
+  - Crawls shared drives in the background and saves them locally
+  - Saves replays of connections so you can look at them later
+  - Runs console commands or PowerShell payloads automatically on new connections
 - RDP Player:
-    - See live RDP connections coming from the MITM
-    - View replays of RDP connections
-    - Take control of active RDP sessions while hiding your actions
-    - List the client's mapped drives and download files from them during active sessions
+  - See live RDP connections coming from the MITM
+  - View replays of RDP connections
+  - Take control of active RDP sessions while hiding your actions
+  - List the client's mapped drives and download files from them during active sessions
 - Converter tool:
-    - Convert RDP replays to videos for easier sharing
-    - Convert RDP replays to a sequence of low-level events serialized in JSON format
-    - Convert PCAPs to replays, videos or JSON events
-    - Convert decrypted PCAPs (L7 PDUs) to replays, videos or JSON events
+  - Convert RDP replays to videos for easier sharing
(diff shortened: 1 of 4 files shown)


yaleman:
Any chance this can be merged, or is the project dead?

Res260:
Hi, project is basically dead. :(

Maintainer changed job so there is no one to maintain anymore cc @obilodeau

yaleman:
That's a shame. 😞

obilodeau:
@yaleman we don't have a good test harness with many clients and servers. I basically need a bunch of VMs to make sure that this change doesn't cause regression anywhere else.

I could trust you and merge to keep things moving but I would like some guarantees. Did you test with Windows desktop 10/11 as a target and did you test with Windows Server as a target?

Thanks!

yaleman:
I've tested it with Windows 10 and a Windows 2022 server and as far as I can tell it works. 

Had to add some build-time dependencies in the Dockerfile so it'd build

README.md
```diff
@@ -514,39 +533,42 @@ In the future there are plans to support other aspects of PyRDP configuration th
...
+
 We developped our own Bettercap module, `rdp.proxy`, to monster-in-the-middle all RDP connections
 on a given LAN. Check out [this document](docs/bettercap-rdp-mitm.md) for more information.
 
 ### Docker Specific Usage Instructions
 
 Since docker restricts the interactions with the host system (filesystem and network), the PyRDP docker image must be run with some parameters depending on your use case. This section documents those parameters.
 
-We refer to the publicly provided docker image but if you [built your own](#building-the-docker-image) replace `gosecure/pyrdp` with the name of your locally built image.
+We refer to the publicly provided docker image but if you [built your own](docs/devel.adoc) replace `gosecure/pyrdp` with the name of your locally built image.
```
obilodeau, line 555: great catch!

Dockerfile
```diff
@@ -6,7 +6,11 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 python3-pip python3-venv \
         # Required to build RLE module
-        build-essential python3-dev
+    build-essential python3-dev pkg-config \
+    libavformat-dev libavcodec-dev libavdevice-dev libavfilter-dev \
+    libswscale-dev libavutil-dev \
+    # minimize image size
+    && rm -rf /var/lib/apt/lists/*
```
obilodeau: Please keep these changes separate... Lots of care goes into producing a minimally sized build and leveraging PyAV pre-built wheels instead of pulling all these dependencies reduce build time and image size.

README.md
obilodeau, line 1: Please refrain from unrelated changes in commits. It is distracting when we review. That markdown was rendering perfectly fine on GitHub.

README.md
obilodeau: because of this one, I will accept all this files' changes

obilodeau (changes requested):
Please revert the docker changes and submit a separate pull request. Check other docker optimization related pull requests to get a sense of what is the proper fix. I don't have time to guide you now.

Thanks!

Dockerfile
yaleman: it won't build without them, so I couldn't build/run it in that case.

yaleman: Also they're not in the final image either way - they're in the compile phase?

obilodeau: True

obilodeau: However, the dependencies also need to be installed on the CI, which is why it is failing on your PR.

obilodeau: I would have loved to keep track of a PyAV version that is distributed as a binary wheel (no need to build) but they are removing old versions quickly...

obilodeau: I am trying to fix the CI here but it is not going smooth: https://github.com/GoSecure/pyrdp/pull/512

obilodeau: Ok, now it looks good. Once all jobs pass, I'll merge and then you can rebase on main here so I see it pass with your new code. Thanks for the patience.

obilodeau: #512 was merged. Please rebase, test again and hopefully the CI will pass.

obilodeau (approved):

obilodeau:
Thanks for your patience!

PR #497, opened 2025-07-26
Added fix(recorder): replace colons in filenames to ensure Windows compatibilty and runs across other platforms as well

Cap-Levi:
Fix: Make filenames compatible with Windows for pyrdp-convert

This PR fixes an issue where PyRDP would crash on Windows due to invalid characters (colons) in filenames generated for recordings.

Colons (`:`) are replaced with underscores (`_`) to ensure the filenames are valid on Windows systems.

Tested on Windows 11 with Python 3.13 Recording now works as expected.


Res260:
There seems to be a failure in the Ubuntu tests

Cap-Levi:
> There seems to be a failure in the Ubuntu tests

So i checked your bash script, you have hard coded the filename file -E "20200319000716_192.168.38.1:20989-192.168.38.1:3389.pyrdp", the whole purpose of this line is to remove this : colon so it can run across any platform, i have ran it on my kali linux, it worked fine on it.

Res260:
No worries, if you could just update the line on the test so it passes it'd be enough!

Cap-Levi:
I have fixed it, you can check now

Cap-Levi:
let me test it on my ubuntu

Cap-Levi:
alright now the integration test is working on my ubuntu. there was one more file where the colon was getting generated now its fixed @Res260 you can run the workflow again

